PR #12, opened 2018-12-19
Boostrap option

GrandSchtroumpf:
Add boostrap option in the AppManager to listen on activation.
- All modules are activated by default.
- You need to activate the plugins (no cache history yet)


src/app-manager.ts
```diff
@@ -1,37 +1,70 @@
...
 
   constructor(dependancies: {
     modules?: { json: ModuleProfile; api: API }[]
     plugins?: { json: PluginProfile; api: Plugin }[]
+    options?: {
+      boostrap: string
+    }
   }) {
     // Modules
     const modules = dependancies.modules || []
```
yann300: there is a typo here, this is `dependencies`

src/app-manager.ts
```diff
@@ -49,31 +82,97 @@ export class AppManager {
...
-  /** Activate a plugin */
+  /** Activate a plugin or module */
   public activate(type: string) {
     if (!this[type]) throw new Error(`Plugin ${type} is not registered yet`)
-    this[type].activate()
-    // TODO : this.addApi()
+    if (this.plugins[type]) {
+      const { json, api } = this.plugins[type]
+      this.activateApi(json, api)
+      this.activatePlugin(json, api)
```
yann300: that one seems to register the notifications the plugin are listening to . Am I right?
In that case that should also potentially apply to modules.
if this is not the case right now, Perhaps better to rename to `registerListenOnNotification` or something like that.

src/app-manager.ts
```diff
@@ -1,37 +1,70 @@
...
-      notifications.forEach(({ type, key }) => {
-        if (!this.events[api.type]) this.events[api.type] = {}
-        if (!this.events[api.type][type]) this.events[api.type][type] = {}
-        this.events[api.type][type][key] = api.notifs[type][key]
-      })
-    })
+  /** Broadcast a message to every plugin listening */
+  private broadcast(type: string, key: string, value: any) {
+    for (const origin in this.events) {
+      if (this.events[origin][type]) {
```
yann300: here the root key is `origin` ( in `this.events[origin][type]` ) .
but line 91, `api.type` is used...

src/extension.ts
```diff
@@ -18,24 +20,62 @@ export class RemixExtension {
...
     if (!msg) return
 
-    const { action, key, type, value } = msg
-    if (action === 'notification') {
-      if (this.notifications[key] && this.notifications[key][type]) {
-        this.notifications[key][type](value)
+    const { action, key, type, value, id, error } = msg
+    try {
+
+      if (action === 'request' && key === 'handshake') {
```
yann300: an `handshake` requires at least 3 messages before the connection can be made no? seems only one is requested here

src/app-manager.ts
GrandSchtroumpf: In the current design, module api can inject other module apis (which enables types on autocompletion and better customization). That's why I didn't put the notification for module api activation. The `activatePlugin` also activates the `request` method of the plugin api. I can call it `activateRequestAndNotification` but it doesn't say that it restricted to plugin api ...
What do you think?

src/app-manager.ts
GrandSchtroumpf: Line 91, the origin of the event triggered is the `api.type`, but I can create a variable to make it easier to understand. Something like: `const origin = api.type`.

src/extension.ts
GrandSchtroumpf: Not sure to understand.
Here the handshake comes from the IDE on plugin activation. It enables the plugin to store the source window without using `parent.window`.
We can also add a handshake response from the plugin to the IDE if it solves a security leak, but I don't see which one right now.
Anyway, it would be maximum 2 messages no ?

src/app-manager.ts
yann300: At least we should make obvious in the code that `api.type` is an URL (is it ?). at first I thought that `api.type` is the reference name of the plugin.

yann300: I guess for module it is not an URL, but there should be some token describing the location of the module.

GrandSchtroumpf: It's not an url (it has nothing to do with event.origin). It's the type of the plugin listening on this event;
Basically if "ethdoc" listen on the method "compilationFinished" of module "solCompiler" we would have : 
- origin : "ethdoc"
- type: "solCompiler"
- key: "compilationFinished"
What word would you use for that?

src/extension.ts
yann300: yes you're right, the third message would be final acknowledge from the parent to the plugin, but that's not mandatory.
Where is the handshake msg from plugin to parent in our case?

GrandSchtroumpf: There is none right now because I don't actually see what security leak it would solve...

src/app-manager.ts
yann300: ok I see, oh `activateRequestAndNotification` looks good to me.
so as i understand, if we want a module to make request to iframe plugin or a module to listen on iframe plugin event, we'll need to call `activatePlugin`, right?

GrandSchtroumpf: I wouldn't recommend a module to make a request to a plugin or listen to plugin events as they are not active but default, and are not maintained by the core team. It should be up to the plugin to send a request to the module. Though two plugins can communicate with each other through events.

src/app-manager.ts
GrandSchtroumpf: Sure. What about `listener`?

yann300: sounds good!

PR #14, opened 2018-12-31
Events

GrandSchtroumpf:
Use the [events library](https://www.npmjs.com/package/events).
Update documentation.
Update examples.
Update Tests.


examples/modules/txlistener.module.ts
```diff
@@ -32,9 +35,9 @@ export class TxlistenerApi extends API<Txlistener> implements Txlistener {
 // External class
 export class TxEmitter {
```
yann300: In the current implementation (remix IDE code), I am not sure we need to have a `TxEmitter` . the current `TxListener` is listening either on new blocks or in a ethereumjs-vm event but it does not need to register directly to an emitter.
but that's a detail, we can merge this API def with the current implementation when it's time

src/app-manager.ts
```diff
@@ -64,26 +72,21 @@ export class AppManager {
...
+  private activateApi<M extends Api>(json: ModuleProfile<M>, api: API<M>) {
     const events = json.events || []
     events.forEach(event => {
-      if (event in api) {
-        api[event].on((value: any) => {
-          this.broadcast(api.type, event, value)
-        })
-      }
+      if (!api.events) return
+      api.events.on(event, (value: any) => this.broadcast(api.type, event, value))
```
yann300: does we need the template here on the call to `broadcast`

src/app-manager.ts
```diff
@@ -130,12 +134,12 @@ export class AppManager {
 
     const methods = json.methods || []
     methods.forEach(key => {
-      if (key in api) delete this[api.type][key]
+      if (key in api) delete this[api.type as string][key]
```
yann300: why do we need `as string` here? `type` is not already enforced to be a string as it is declared in the interface `API`?

examples/modules/txlistener.module.ts
GrandSchtroumpf: It's just an example of how a module can deal with an external class. In the end the implementation won't look exactly like that.

src/app-manager.ts
GrandSchtroumpf: What do you mean by "template" ?

src/app-manager.ts
GrandSchtroumpf: In this case, it's kind of a mismatch between the generic Api "M" in this method and what the `this.calls` method accept with the generic IAppManager "T" in the class. Removing the `as string` create a typescript warning. It's a quite fix to remove a Typescript warning, it won't impact the end Javascript code. But yes, I should improve that in time.

src/types.ts
```diff
@@ -1,24 +1,42 @@
...
-  events?: (Extract<keyof T, string>)[]
+  methods?: ExtractKey<T, Function>[]
+  events?: (keyof T['events'])[]
 }
 
-export interface PluginProfile {
-  type: string
-  methods?: string[]
-  events?: string[]
+export interface PluginProfile<T extends Api = any> extends ModuleProfile<T> {
```
yann300: `PluginProfile` refers to a plugin JSON declaration right?
why removing `type`, `methods`, and `events`?

GrandSchtroumpf: I'm not, it's extending the `ModuleProfile` instead which has the `type`, `methods` and `events`.

PR #22, opened 2019-03-08
Api factory

GrandSchtroumpf:
Do not merge yet 💣
Introduce an API Factory to manage request.
It aims to merge the Profile and Api interface.


src/extension-client/remix-extension.ts
```diff
@@ -117,7 +114,7 @@ export class RemixExtension<T extends Api = any> {
   /** Return the current theme when handshaked */
   public loaded(): Promise<{theme: string}> {
```
yann300: I understand that this it not the current focus for this PR but:
seems that if `loaded()` is called after the plugin is actually loaded, the Promise is never resolved.

yann300: for `/** Return the current theme when handshaked */` as we discussed, that is better to have a `config` plugin.

src/engine/api-factory.ts
```diff
@@ -0,0 +1,62 @@
...
+    return {
+      api,
+      ...this.profile,
+      addRequest: (request: PluginRequest, method: string, args: any[]) => this.addRequest(request, method, args)
+    }
+  }
+}
+
+
+class AppManager {
```
yann300: that is temporary code right?

GrandSchtroumpf: yes

src/extension-client/remix-extension.ts
GrandSchtroumpf: > the Promise is never resolved.

This is not the aim of this method to be called after the plugin is loaded. It's like if you start listening on `body.onload` once the body is loaded... But we can have a variable `isLoaded` is needed.

>  that is better to have a config plugin.

Yes, I'll remove that

src/engine/app-manager.ts
```diff
@@ -111,57 +112,57 @@ export abstract class AppManagerApi implements API<AppManager> {
...
-    this.events.emit('activate', entry)
+    this.events.emit('activate', api)
   }
 
   /** Activation for Module and Plugin */
-  private activateCallAndEvent<T extends Api>({ profile, api }: Entry<T>) {
-    const events = profile.events || []
-    events.forEach(event => {
+  private activateCallAndEvent<T extends Api>(api: PluginApi<T>) {
+    const events = api.profile.events || []
```
yann300: should be `api.events` ?

GrandSchtroumpf: `api.events` : the event emitter
`api.profile.events` : the names of the events exposed

GrandSchtroumpf:
Should be ready to merge.

src/engine/app-manager.ts
```diff
@@ -111,57 +112,54 @@ export abstract class AppManagerApi implements API<AppManager> {
...
       api.events.on(event, (...payload: any[]) =>
-        this.broadcast(profile.name, event as string, payload),
+        this.broadcast(api.name, event as string, payload),
       )
     })
 
-    const methods = profile.methods || []
-    this.calls[profile.name] = {}
+    const methods = api.profile.methods || []
+    this.calls[api.name] = {}
```
yann300: is this `api.profile.name` ?

src/engine/plugin.ts
```diff
@@ -19,6 +29,7 @@ export class Plugin<T extends Api> {
   } = {}
 
   public readonly name: T['name']
+  public profile: PluginProfile<T>
   public events: ApiEventEmitter<T>
   public notifs = {}
   public request: (value: { name: string; key: string; payload: any }) => any
```
yann300: should `any` be specified to be a promise here?

src/engine/app-manager.ts
GrandSchtroumpf: `api.name === api.profile.name`. I use that to make it easier to filter in a store. Both values are `readonly` so there is no risk that the value changes.

PR #36, opened 2019-03-13
allow dev mode

yann300:
fixes https://github.com/ethereum/remix-plugin/issues/29


src/extension-client/remix-extension.ts
```diff
@@ -14,7 +14,8 @@ export class RemixExtension<T extends Api = any> {
   private id = 0
   private handshake: () => any
   protected currentRequest: PluginRequest
-  public isLoaded = false
+  public isLoaded: boolean = false
```
GrandSchtroumpf: It's best practice to not define the type when it's implied : 
```
public isLoaded = false
public devMode = false
```

PR #38, opened 2019-03-14
Render hook

GrandSchtroumpf:
Add a hook `render` instead of `activate` & `create` in `Plugin`.

### DIFF
--- a/src/engine/api-factory.ts
+++ b/src/engine/api-factory.ts
@@ -10,6 +10,7 @@ import {
 export abstract class ApiFactory<T extends Api = any> {
   abstract readonly profile: ModuleProfile<T>
   abstract events?: ApiEventEmitter<T>
+  abstract render?: () => HTMLElement
   private requestQueue: Array<() => Promise<any>> = []
   protected currentRequest: PluginRequest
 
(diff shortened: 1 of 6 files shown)


src/engine/api-factory.ts
```diff
@@ -18,6 +19,7 @@ export abstract class ApiFactory<T extends Api = any> {
       events: this.events,
       name: this.profile.name,
       profile: this.profile,
+      render: this.render ? () => this.render() : undefined,
```
yann300: i m not quite sure about the syntax here... is the `return` keyword  missing here?

GrandSchtroumpf: `render: this.render ? () => this.render() : undefined,` is the same as : 
```javascript
render: () => {
  if (!!this.render) {
    return () => { return this.render() }
  } else {
    return undefined
  }
}
```

src/engine/plugin.ts
```diff
@@ -147,57 +137,33 @@ export class Plugin<T extends Api> implements PluginApi<T> {
...
-        this.postMessage({
-          action: 'request',
-          name: this.name,
-          key: 'handshake',
-        })
-      }
-    } catch (err) {
-      console.log(err)
+  public render() {
+    if (this.iframe) {
```
yann300: that might be a problem if one activate / deactivate / activate a plugin, cause the instance seems not reset upon deactivation. Perhaps should we keep a `deactivation` hook for reset this object

yann300: ah i see `deactivate`, should the iframe ref should be removed to

GrandSchtroumpf: Yes, indeed. Thanks

PR #42, opened 2019-03-15
Improve error handling

GrandSchtroumpf:
Use `setDevMode` to specify a port for the IDE
Improve error message when source is undefined.

### DIFF
--- a/src/extension-client/remix-extension.ts
+++ b/src/extension-client/remix-extension.ts
@@ -1,7 +1,11 @@
 import { Message, Api, PluginRequest } from '../types'
 
+interface DevMode {
+  port: number | string
+}
+
 export class RemixExtension<T extends Api = any> {
-  private devModePort: number | string
+  private devMode: DevMode
   private source: Window
   private origin: string
   private notifications: {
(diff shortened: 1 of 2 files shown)


src/extension-client/remix-extension.ts
```diff
@@ -15,7 +16,6 @@ export class RemixExtension<T extends Api = any> {
   private handshake: () => any
   protected currentRequest: PluginRequest
   public isLoaded = false
-  public devMode = false
```
yann300: could we keep `devMode`? as we might need to use it for something else than filtering out the right dev port.

GrandSchtroumpf: Then we can just check is `devModePort` exists ?
Maybe we can have a default `8080` value in the `setDevMode` function in case someone forget to fill the argument : 
```javascript
public setDevMode(devModePort: number | string = 8080) {
  this.devModePort = devModePort
}
```

yann300: > Then we can just check is devModePort exists ?

`devMode` support could at some point involve other property than `devModePort`. 
That's why a global switch `devMode` could be handy

GrandSchtroumpf: We can have an object `devMode` with a property `port` if you think it's better. But we should be careful not to have too much one-property-object that linger in the code.

PR #43, opened 2019-03-18
WIP : Create permission handler

GrandSchtroumpf:
💣 WIP : Don't merge yet .
Create a Permission Handler to ask the user his permission for a plugin to call another one.


src/engine/permission-handler.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (!this.permissions[to]) this.permissions[to] = {}
+    // Never allowed
+    if (!this.permissions[to][from.name]) {
+      const { allow, remember } = await this.openPermission(from, to, false)
+      if (remember) this.permissions[to][from.name] = { allow, hash: this.permissions[to][from.name].hash }
+      this.setPermissions()
+      return allow
+    }
+    // Remember not allow
+    if (!this.permissions[to][from.name].allow) {
```
yann300: should this return `allow` if hash is hash has not changed?

src/engine/permission-handler.ts
```diff
@@ -0,0 +1,86 @@
...
+      allow: boolean,
+      hash: string,
+    }
+  }
+}
+
+export interface IPermissionProvider {
+  confirm(message: string): Promise<{ allow: boolean, remember: boolean }>
+  /** Clear the permission object */
+  clear(): void
```
yann300: isn't this part should be in the `permissionHandler` interface? (cause this is directly modifying the permissions list

examples/modules/app-manager.api.ts
```diff
@@ -1,6 +1,8 @@
 import { AppManagerApi, Api, PluginApi } from '../../src'
+import { SecurityHandler } from './permission-handler'
 
 export class RemixAppManager extends AppManagerApi {
+  public permissionHandler = new SecurityHandler()
```
yann300: naming it `securityHandler` would make more sense I suppose?

src/engine/permission-handler.ts
```diff
@@ -0,0 +1,86 @@
...
+   * @param wasAllow Did the plugin have changed its hash
+   */
+  private async openPermission(
+    from: {name: string, hash: string},
+    to: string, wasAllow: boolean
+  ): Promise<{ allow: boolean, remember: boolean }> {
+    const msg = wasAllow
+      ? `${from.name} would like to access plugin ${to}.\n Check it's new content ${from.hash}`
+      : `${from.name} has changed and would like to access the plugin ${to}.\n Check it's new content ${from.hash}`
+    return this.confirm(msg)
```
yann300: could we pass in profile of both plugin / module ? so we can display more information if needed?

examples/modules/app-manager.api.ts
GrandSchtroumpf: The abstract class is `PermisionHandler` that's why the attribute is called `permissionHandler`. Here `SecurityHandler` extends `PermissionHandler`, so it is _a kind of_ `PermissionHandler`.

> also I think this SecurityHandler should be a param of appManager right?

When you say "param", you mean a parameter of the constructor ? Right now it's an abstract attribute that needs to be implemented. But I'm open to other options.

src/engine/permission-handler.ts
GrandSchtroumpf: Yes, I think `clear` should be in the Handler. I'll change that.

src/engine/permission-handler.ts
GrandSchtroumpf: Sure

src/engine/permission-handler.ts
GrandSchtroumpf: We can, but in this case, if the user sets "remember, no-permission", then if the hash changes, it would show the permission modal again...

examples/modules/app-manager.api.ts
yann300: > you mean a parameter of the constructor 
 
yes

> it's an abstract attribute that needs to be implemented

so there's a `set` function for this?

GrandSchtroumpf: Hmmm, not sure to understand. I'll clean the code a little, and add more examples. Hopefully it's going to be easier to understand.

examples/permission/asbtract-class.ts
```diff
@@ -0,0 +1,87 @@
...
+  abstract confirm(
+    message: string,
+  ): Promise<{ allow: boolean; remember: boolean }>
+
+  constructor() {
+    const permission = localStorage.getItem('permissions')
+    this.permissions = permission ? JSON.parse(permission) : {}
+  }
+
+  private setPermissions() {
```
yann300: could we have a different naming here?
like `persistPermissions`

examples/modules/app-manager.api.ts
yann300: i wasn't sure how you plan to be able to set a `PermissionHandler` (by having a `setPermissionHandler` or directly in the constructor) . But I think your last commits make it clear

GrandSchtroumpf: I think every method is ok, either in the constructor, or as a parameter, or set later with `app.permissionHandler = new PermissionHandler()`. It shoud work for all.

examples/permission/asbtract-class.ts
GrandSchtroumpf: yes, this is an example. You don't actually need to implement this method. I need it here to avoid having a getter/setter directly on `localStorage`. I'll change it in the example to make it clearer to understand though. Thanks.

PR #44, opened 2019-03-20
make remix extension listen on theme changed

yann300:


yann300:
that's a proposal for ensuring plugin used same bootstrap as the main app

src/extension-client/remix-extension.ts
```diff
@@ -1,5 +1,9 @@
 import { Message, Api, PluginRequest } from '../types'
 
+export interface RemixExtensionOptions {
+  useCustomBootStrapTheme: false
```
GrandSchtroumpf: Define the type of `useCustomBootStrapTheme` as a `boolean`. Default value should be managed by an external method or in the constructor.

src/extension-client/remix-extension.ts
```diff
@@ -16,8 +20,20 @@ export class RemixExtension<T extends Api = any> {
   protected currentRequest: PluginRequest
   public isLoaded = false
 
-  constructor() {
+  constructor(opts: RemixExtensionOptions) {
```
GrandSchtroumpf: Change `opts` to `options` if it's ok. It's more readable for language service tools.
Also make it optional: `constructor(options?: RemixExtensionOptions`.

src/extension-client/remix-extension.ts
```diff
@@ -16,8 +20,20 @@ export class RemixExtension<T extends Api = any> {
...
     window.addEventListener('message', event => this.getMessage(event), false)
+
```
GrandSchtroumpf: Maybe create a `defaultOptions(options)` method to define the default values : 
```typescript
function defaultOptions(options? : Partial<RemixExtensionOptions>): RemixExtensionOptions {
  const opts = options || {}
  return {
    useCustomBootStrapTheme: opts.useCustomBootStrapTheme || false
  } as RemixExtensionOptions
}
```
and use it in the constructor :
```typescript
const opts = defaultOptions(options)
this.initListenOnSwitchTheme(opts.useCustomBootStrapTheme)
```

src/extension-client/remix-extension.ts
```diff
@@ -16,8 +20,20 @@ export class RemixExtension<T extends Api = any> {
...
+    if (!opts.useCustomBootStrapTheme) this.initListenOnSwitchTheme()
+  }
+
+  private initListenOnSwitchTheme () {
```
GrandSchtroumpf: To avoid having to many `if` statement in the constructor we can have the `if` in the method : 
```typescript
private initListenOnSwitchTheme (hasCustomTheme: boolean) {
  if (hasCustomTheme) return
  //...
}
```

src/extension-client/remix-extension.ts
```diff
@@ -16,8 +20,20 @@ export class RemixExtension<T extends Api = any> {
...
+    let cssLink = document.createElement('link')
```
GrandSchtroumpf: Here `cssLink` is never reassigned, use `const` instead of `let`

src/extension-client/remix-extension.ts
```diff
@@ -16,8 +20,20 @@ export class RemixExtension<T extends Api = any> {
...
+    cssLink.setAttribute('rel', 'stylesheet')
+    document.head.appendChild(cssLink)
+    this.listen('theme', 'switchTheme', (theme) => {
```
GrandSchtroumpf: Type `theme` if you know the properties : `(theme: {url: string, quality: 'dark' | 'light'})`

src/extension-client/remix-extension.ts
yann300: wouldn't defining this in the ctor work:
`constructor(options: RemixExtensionOptions = { useCustomBootStrapTheme: false }) {`

PR #54, opened 2019-03-28
Add documentation for status

GrandSchtroumpf:


readme.md
```diff
@@ -206,4 +207,23 @@ newTransaction
 extension.listen('txlistener', 'newTransaction', (tx: Object) => {
   // Do Something
 })
+```
+
+
+# Status
+Every plugin has a status object that can display notifications on the IDE. You can listen on a change of status from any plugin using `statusChanged` event :
```
yann300: perhaps should we also mention that any plugin can update its own status. using `setStatus(...)` from remix-extension?

GrandSchtroumpf: Can we ? Is it implemented yet ?

PR #58, opened 2019-04-01
Issue#57 improve apifactory status

GrandSchtroumpf:
Create two stores that manage the state of the module : 
- `Store`: For a custom object state.
- `EntityStore`: For a state that has a collection of objects.

The Store inherits the `ApiFactory`.


src/stores/store.ts
```diff
@@ -0,0 +1,91 @@
...
+/** Create an empty state */
+function createState<T extends State>(): T {
+  return {
+    status: {} as Status,
+  } as T
+}
+
+export const StoreProfile: Partial<ModuleProfile> = {
+  events: ['statusChanged'],
+  notifications: {}
```
yann300: should this include 
```

 {
 'theme': ['switchTheme']
 }


```

GrandSchtroumpf: I we merge `Store` and `ApiFactory` yes. Right now it's in the `defaultProfile`...

yann300: then we should probably merge them

src/api/status.ts
```diff
@@ -0,0 +1,37 @@
+import { Status, Api, ApiEventEmitter, ModuleProfile } from "../types"
+import { BaseMixinApi, MixinApi } from "./mixin"
+
+// API
+export interface StatusApi {
+  name: 'status',
+  events: {
```
yann300: should the function `getStatus` be  declared here?

src/api/base.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export const baseProfile = {
+  notifications: {
+    'theme': ['switchTheme']
+  }
+}
+
+export abstract class BaseApi<S, U extends Api> {
+  private initialState: S
+  protected state: S
```
yann300: why do you keep `initialState` here?

yann300: nevermind
https://github.com/ethereum/remix-plugin/pull/58/files#diff-74ebb899b68ffa95e80a782b8980aa1bR96

src/api/base.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.state = this.initialState
+  }
+
+
+  /////////
+  // API //
+  /////////
+
+  /** Exports an Api interface */
+  public api(): PluginApi<U> {
```
yann300: just a thought: should we change the name `PluginApi` at some point? because it seems to refer to a Plugin, although it is also valid for Module

src/api/status.ts
GrandSchtroumpf: Sure, I'll add it.

src/api/base.ts
GrandSchtroumpf: Yes we should. Any idea ? =)

yann300:
`api-factory` can be removed right?

src/api/entity.ts
```diff
@@ -0,0 +1,195 @@
...
+function EntityState<T>(): EntityState<T> {
+  return {
+    ids: [],
+    actives: [],
+    entities: {},
+  }
+}
+
+// PROFILE
+export const entityStoreProfile: Partial<ModuleProfile<EntityApi<Api>>> = {
```
yann300: that is not used at the moment?

GrandSchtroumpf: No, I'll use it in the example for the e2e test for `FileManager` though

src/api/base.ts
yann300: `RuntimeAPI` maybe (but perhaps not in this PR)

yann300:
how module (I mean compile module from Remix IDE for instance) should use inheritance with these changes?
look like they have to extends `extends MixinApi([StatusMixin])<StatusState, T> implements PluginApi<T> {` if they want to emit / use status API

GrandSchtroumpf:
For Status : 
```javascript
class Compiler extends MixinApi([StatusMixin]) {
 constructor() {
   super(createProfile(compilerProfile, statusProfile)
 }
}
```
For Status and Entity :
```javascript
class FileManager extends MixinApi([StatusMixin, EntityMixin]) {
 constructor() {
   super(createProfile(fileManagerProfile, statusProfile, entityMixin)
 }
}
```
Not the best devX though. I'd like to improve it. But I think I need to finish this PR before.
What do you think ?

yann300:
the extension of `profile` `super(createProfile(fileManagerProfile, statusProfile, entityMixin)
 }` should be done in the base class.

The thing is that i don't think I agree the fact that the `MixinApi` contain the `BaseAPI`. looks like we are too much depending on a possible `state`.

GrandSchtroumpf:
I can add the `BaseApi` as first argument of the `MixinApi` like that we don't need the `state` : 
```javascript
export function MixinApi(Base, mixins: any[]) {
  mixins.forEach(mixin => {
    Object
      .getOwnPropertyNames(mixin.prototype)
      .forEach(name => Base.prototype[name] = mixin.prototype[name])
  })
  return Base
}
```
But the idea of this PR is to manage the `status` in all plugin/module the same way... If we don't make the `state` mandatory, then there is not much interest in those changes.

GrandSchtroumpf:
> the extension of profile super(createProfile(fileManagerProfile, statusProfile, entityMixin) } should be done in the base class.

It's not possible because it depends on the Mixin we want to add. And we cannot do it int the `MixinApi` because we don't have access to the attributes in the `mixin.prototype`...

GrandSchtroumpf:
Maybe it's just not the right way of doing. Let's talk about the design together on gitter.

src/api/compiler.ts
```diff
@@ -0,0 +1,24 @@
+import { BaseApi, extendsProfile } from "./base"
+import { ModuleProfile, Api, API } from "src/types"
+
+export interface ICompilerApi extends Api {
+  events: {
+    compilationFinished: []
```
yann300: is it possible to define the parameters here: 
`file: string, source: {sources:<map of files>, target: string}} , languageVersion: string, data: compilationResult ` ?
target is the same as file and compilationResult should be https://solidity.readthedocs.io/en/v0.5.7/using-the-compiler.html#output-description
(but perhaps a bit too much to type all that)

GrandSchtroumpf: Sure ! This is the goal of this interface. I just didn't know where to find the types. Thanks for the documentation.

Issue #63, opened 2019-04-08
Api for FileManager


GrandSchtroumpf:
Add the Api interface for FileManager

PR #64, opened 2019-04-09
Issue#63 api file manager

GrandSchtroumpf:
Fixes https://github.com/ethereum/remix-plugin/issues/63
Improve types for better autocompletion on Typescript


src/api/udapp/api.ts
```diff
@@ -4,20 +4,21 @@ import { RemixTx, RemixTxReceipt } from './type'
 
 export interface IUdappApi extends Api {
   events: {
-    newTransaction: [RemixTx]
+    newTransaction: (transaction: RemixTx) => void
```
yann300: i think `newTransaction` also return the TxReceip as second params

yann300: oups scratch that please ;)

Issue #67, opened 2019-04-12
Add local api access for PluginClient


GrandSchtroumpf:
The PluginClient should have access to the api without `listen` and `call`.
- [X] Split the class into function.
- [X] Make recursive type definition for the api creation.
- [ ] Create a method to expose some methods to other plugins
- [ ] Test each function.

PR #68, opened 2019-04-12
Issue#67 pluginclient api

GrandSchtroumpf:
**Ready to merge**
Fix issue https://github.com/ethereum/remix-plugin/issues/67


src/store/base.ts
```diff
@@ -0,0 +1,47 @@
+// BASE
+export class BaseStore<S extends {}> {
```
yann300: seems not used anymore - shall we remove it?

GrandSchtroumpf: Yes right.

src/client/iframe.ts
```diff
@@ -0,0 +1,101 @@
...
+          event.source.postMessage(message, event.origin as any)
+        })
+      }
+
+      // Check if is loaded
+      if (!loaded) throw new Error('Handshake before communicating')
+
+      switch (action) {
+        case 'notification': {
+          client.events.emit(`[${name}] ${key}`, ...payload)
```
yann300: interesting way of providing separation

yann300: indeed

src/client/iframe.ts
```diff
@@ -0,0 +1,101 @@
...
+    "https://remix.ethereum.org",
+    ...localhost
+  ].includes(msgOrigin)
+}
+
+/**
+ * Start listening on the IDE though PostMessage
+ * @param store A store to put the messages into
+ */
+function connectIframe(client: PluginClient) {
```
yann300: as a first tentative, we could try to implement function `connectWS(client: PluginClient) {`

src/client/iframe.ts
yann300: should that one use https://github.com/ethereum/remix-plugin/pull/68/files#diff-2609dc589ff34d954b51407ff2140828R32 ?

PR #72, opened 2019-04-29
Plugin example

GrandSchtroumpf:
**Ready to merge**
Create an example of plugin for early plugin developers
- [X] Listen on compiler
- [X] Call the FileSystem
- [x] Use Bootstrap style


examples/plugins/ethdoc/src/ethdoc.ts
```diff
@@ -0,0 +1,115 @@
+import {
+  CompilationResult,
+  CompilatedContract,
```
yann300: there's a little typo here, this is `compiledContract`

examples/plugins/ethdoc/src/ethdoc.ts
```diff
@@ -0,0 +1,115 @@
...
+  DeveloperDocumentation,
```
yann300: All these types are not really related to `remix-plugin` , we could easily have an another NPM package for referencing all the solidity related types ( I recall that something we already being discussed by the solidity team right?)

src/api/compiler/type/input.ts
```diff
@@ -0,0 +1,117 @@
...
+  /** disabled by default */
+  enable: boolean
+  /**
+   * Optimize for how many times you intend to run the code.
+   * Lower values will optimize more for initial deployment cost, higher values will optimize more for high-frequency usage.
+   */
+  runs: number
+}
+
+export interface CompilaterMetadata {
```
yann300: typo `CompiledMetadata`

examples/plugins/ethdoc/src/ethdoc.ts
GrandSchtroumpf: We were thinking about putting them into the solc-js repository. But I'll need to change several things to their repo before. So it might take more time.

src/api/compiler/type/input.ts
```diff
@@ -0,0 +1,117 @@
...
+    | 'spuriousDragon'
+    | 'byzantium'
+    | 'constantinople'
+  /** Metadata settings */
+  metadata?: CompilaterMetadata
+  /** Addresses of the libraries. If not all libraries are given here, it can result in unlinked objects whose output data is different. */
+  libraries: CompilerLibrarie
+}
+
+export interface CompilaterOptimizer {
```
yann300: typo `CompiledOptimizer`

src/api/compiler/type/input.ts
```diff
@@ -0,0 +1,117 @@
...
+  evmVersion:
+    | 'homestead'
+    | 'tangerineWhistle'
+    | 'spuriousDragon'
+    | 'byzantium'
+    | 'constantinople'
+  /** Metadata settings */
+  metadata?: CompilaterMetadata
+  /** Addresses of the libraries. If not all libraries are given here, it can result in unlinked objects whose output data is different. */
+  libraries: CompilerLibrarie
```
yann300: typo `CompiledLibraries`

src/api/compiler/type/input.ts
GrandSchtroumpf: thx

PR #88, opened 2019-06-04
Add typings for Plugins and Engine

GrandSchtroumpf:
Improve typings for typings and Engine.
Make the permission handler optional.
fix some issues.


yann300:
+1

PR #96, opened 2019-06-21
Add status interface

### LINKED ISSUE #95: Add status changes inside Api
All plugins should have a status by default.

GrandSchtroumpf:
Fix #95

### DIFF
--- a/projects/utils/src/api/editor/api.ts
+++ b/projects/utils/src/api/editor/api.ts
@@ -2,7 +2,7 @@ import { HighlightPosition } from './type'
 import { StatusEvents } from '../../types'
 
 export interface IEditor {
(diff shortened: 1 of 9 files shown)


projects/utils/src/api/editor/api.ts
```diff
@@ -1,7 +1,8 @@
 import { HighlightPosition } from './type'
+import { StatusEvents } from '../../types'
 
 export interface IEditor {
-  events: {}
+  events: StatusEvents
```
yann300: Tiny comment, for better readability perhaps we should always write `{} & StatusEvent`
If no other event are described. Also I suppose `StatusEvent` should be singular.

PR #109, opened 2019-07-16
Methods on handshake

GrandSchtroumpf:
Plugin can pass its methods on handshake

### DIFF
--- a/projects/engine/src/engine/plugin-engine.ts
+++ b/projects/engine/src/engine/plugin-engine.ts
@@ -83,10 +83,6 @@ export class PluginEngine<T extends ApiMap> extends AbstractPluginEngine {
     this.register(Object.keys(plugins).map(key => plugins[key]))
   }
 
-  // public get actives(): string[] {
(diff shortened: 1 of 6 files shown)


projects/engine/src/engine/plugin-engine.ts
```diff
@@ -82,9 +83,9 @@ export class PluginEngine<T extends ApiMap> extends AbstractPluginEngine {
     this.register(Object.keys(plugins).map(key => plugins[key]))
   }
 
-  public get actives(): string[] {
-    return Object.keys(this.methods)
-  }
+  // public get actives(): string[] {
```
yann300: shouldn't that stay?

yann300:
from that, does retrieving the profile from here, also return the newly added methods?

projects/engine/src/engine/plugin-engine.ts
GrandSchtroumpf: Ah I forgot to remove the comments. No it's defined above with `protected actives: string[] = []`

PR #114, opened 2019-07-23
Plugin onload

GrandSchtroumpf:
Update Ethdoc and fix onload callback bug


yann300:
if the changes in this PR ensure it s working for both remix and remix-alpha, i m fine with that

PR #116, opened 2019-07-26
Remixide types

GrandSchtroumpf:
Add types for solidityUnitTesting and contentImport plugins.


yann300:
Is that associated with a PR in remix IDE?

GrandSchtroumpf:
No, it's just to have typings available on the iframe side.

yann300:
Yes , i just meant that the function signature in Remix IDE is not the same then (for unitesting module)

GrandSchtroumpf:
Oh, right, I'll do a PR. Thanks

PR #120, opened 2019-08-12
Add "allow-popups" to iframe attributes

pi0neerpat:
I'm building a plugin which uses external links for the user to open.

## Current state for plugins/iframe
The iframe attributes currently include `allow-top-navigation`. This allows the user to open a link, but redirects the page, and closes Remix-IDE :-1: 

## Changes
This change to add "allow-popups" means the user can click a link without leaving Remix-IDE. :+1: 

Thanks for making the plugin process easy :zap: yall rock :guitar: , and I've had a lot of fun :100:

### DIFF
--- a/projects/engine/src/plugin/iframe.ts
+++ b/projects/engine/src/plugin/iframe.ts
@@ -103,7 +103,7 @@ export class IframePlugin extends ViewPlugin {
     if (this.iframe.contentWindow) {
       throw new Error(`${this.name} plugin is already rendered`)
     }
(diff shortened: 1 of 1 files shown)


pi0neerpat:
Just a dev-ops question. How often is the plugin updated on remix and remix-alpha? This PR doesn't seem to be applied there yet.

GrandSchtroumpf:
Hi, sorry for the delay, I'll update remix-ide right now. Should be available on remix-alpha in an hour or two.

pi0neerpat:
Awesome thanks for the update!

PR #125, opened 2019-08-19
Clarifies that sendTransaction is VM only, udapp call fix

bmcd87:
Looking in remix-lib/src/universalDapp.js, sendTransaction() looks like it only disallows mainnet transactions, but it calls silentRunTx() which does an additional check that makes sure you are using javascript VM. Updating readme to clarify this.

Also one of the example had a copy-paste error, calling client.solidity.on instead of client.udapp.on.

### DIFF
--- a/doc/plugins/udapp.md
+++ b/doc/plugins/udapp.md
@@ -8,7 +8,7 @@ The udapp exposes an interface for interacting with the account and transaction.
 |Type     |Name                 |Description
 |---------|---------------------|--
 |_event_  |`newTransaction`     |Triggered when a new transaction has been sent.
-|_method_ |`sendTransaction`    |Send a transaction **only for testing networks**.
+|_method_ |`sendTransaction`    |Send a transaction **only for JavaScript VM**.
 |_method_ |`getAccounts`        |Get an array with the accounts exposed.
 |_method_ |`createVMAccount`    |Add an account if using the VM provider. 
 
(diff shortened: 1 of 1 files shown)


yann300:
could you rollback this change https://github.com/ethereum/remix-plugin/pull/125/files#diff-1ba3695f6ad726e80e769e44aa0c42a9R11
we will push https://github.com/ethereum/remix/pull/1265

doc/plugins/udapp.md
```diff
@@ -17,7 +17,7 @@ The udapp exposes an interface for interacting with the account and transaction.
 ### Events
 `newTransaction`: Triggered when a new transaction has been sent.
 ```typescript
-client.solidity.on('newTransaction', (tx: RemixTx) => {
+client.udapp.on('newTransaction', (tx: RemixTx) => {
```
GrandSchtroumpf: 👍

doc/plugins/udapp.md
```diff
@@ -8,7 +8,7 @@ The udapp exposes an interface for interacting with the account and transaction.
 |Type     |Name                 |Description
 |---------|---------------------|--
 |_event_  |`newTransaction`     |Triggered when a new transaction has been sent.
-|_method_ |`sendTransaction`    |Send a transaction **only for testing networks**.
+|_method_ |`sendTransaction`    |Send a transaction **only for JavaScript VM**.
```
GrandSchtroumpf: We'll update the remix-ide to support testing networks. Could you rollback this change ?
Same below.

bmcd87: Updated so it only includes the solidity -> udapp change (and i guess a newline character at the end of the file that github added on its own)

PR #134, opened 2019-09-13
Add `fromNative` to the Request

yann300:
linked to https://github.com/ethereum/remix-ide/pull/2330
please feel free to propose a better name for the property than `fromNative`


projects/engine/src/engine/plugin-engine.ts
```diff
@@ -193,7 +193,7 @@ export class PluginEngine<T extends ApiMap> extends AbstractPluginEngine {
         throw new Error(`${notExposedMsg} ${exposedMethodsMsg}`)
       }
 
-      const request = { from: name }
+      const request = { from: name, fromNative: this.settings.native.includes(name)}
```
GrandSchtroumpf: can this be `isNative` or `isFromNative` instead, as this is a boolean value.

PR #137, opened 2019-09-17
Fix defaultGateway

yann300:


projects/engine/src/plugin/util.ts
```diff
@@ -1,9 +1,10 @@
 
-export let defaultGateways: {
+const defaultGateways = {
```
GrandSchtroumpf: export direcly like that : 
```typescript
export const defaulGateways = { ... }
```

PR #151, opened 2019-11-14
[WIP] Adding version control (git) support to remix plugin

edisinovcic:
Created profile for version control (git) on remix plugin
Only a subset of commands is currently supported.


projects/utils/src/api/git/api.ts
```diff
@@ -0,0 +1,35 @@
+import { StatusEvents } from '../../types'
+
+export interface IGitSystem {
+  events: {
+    //TODO:
```
GrandSchtroumpf: As there is no events expected in a git plugin, you can remove TODO

projects/utils/src/api/git/profile.ts
```diff
@@ -0,0 +1,8 @@
+import { IGitSystem } from './api'
+import { LibraryProfile } from '../../types'
+
+export const gitProfile: LibraryProfile<IGitSystem> = {
+  name: 'git',
+  methods: ['clone', 'checkout', 'init', 'add', 'commit', 'fetch', 'pull', 'push', 'reset', 'status', 'remote', 'log']
+  //events: ['currentFileChanged']
```
GrandSchtroumpf: As there is no events expected in a git plugin, you can remove the comment

projects/utils/src/api/git/api.ts
edisinovcic: This has been deleted

projects/utils/src/api/git/profile.ts
edisinovcic: This has also been removed

PR #157, opened 2020-01-18
Adds id attribute to plugin iframes for easier testing

bmcd87:
I've been trying to add integration tests to my plugin, but using nightwatch you need to explicitly step into an iframe using .frame(). This would be a lot easier if we have a unique id to reference, so I added that using the plugin's name (which must be unique). Please let me know what you think.


projects/engine/src/plugin/iframe.ts
```diff
@@ -116,6 +116,7 @@ export class IframePlugin extends ViewPlugin {
     }
     this.iframe.setAttribute('sandbox', 'allow-popups allow-scripts allow-same-origin allow-forms allow-top-navigation')
     this.iframe.setAttribute('seamless', 'true')
+    this.iframe.setAttribute('id', `frame-${this.name}`)
```
GrandSchtroumpf: Maybe `plugin-${this.name}` instead ?

PR #158, opened 2020-01-27
Doc for Plugin Service

GrandSchtroumpf:
Write documentation for Plugin Services.
Write integration test for Plugin Services.
Solve issues found in tests


projects/utils/src/types/api.ts
```diff
@@ -24,7 +24,7 @@ export type MethodParams<T extends Api, K extends MethodKey<T>> = T extends Api
   ? Parameters<T['methods'][K]>
   : any[]
 
-/////////
+  /////////
```
LianaHus: :)

PR #164, opened 2020-02-10
Add API documentation for PluginManager

GrandSchtroumpf:


doc/engine/api/manager.md
```diff
@@ -0,0 +1,254 @@
...
+const profile = { name: 'manager', methods: [...managerMethods, 'isFromRemiw' ] }
+const manager = new RemixManager(profile)
+```
+
+## Properties
+
+### requestFrom
+Return the name of the caller. If no request provided, this mean that the method has been called from the IDE so we use "manager". 
+
+_Use this method when you expose custom methods from the Plugin Manager._
```
LianaHus: you forgot the method

projects/engine/src/engine/next-gen-engine.ts
```diff
@@ -126,12 +126,6 @@ export class Engine {
       this.manager.getProfile(caller),
     ])
 
-    // Check if plugin FROM can call METHOD of plugin TO
```
yann300: that is removed?

GrandSchtroumpf: Yes, as discussed in Gitter, most of the calls won't require any permission at the method level. It's easier to maintain and more flexible to let plugin developers defined when to ask for permission with `canCall`.

PR #171, opened 2020-03-05
Debug plugin interaction

GrandSchtroumpf:
Activate manager from engine
Update canCall

### DIFF
--- a/projects/engine/src/engine/engine.ts
+++ b/projects/engine/src/engine/engine.ts
@@ -14,9 +14,9 @@ export class Engine {
   constructor(private manager: BasePluginManager) {
     this.plugins['manager'] = manager
     // Activate the Engine & start listening on activation and deactivation
-    this.activatePlugin('manager').then(() => {
-      this.manager['engineActivatePlugin'] = (name: string) => this.activatePlugin(name)
-      this.manager['engineDeactivatePlugin'] = (name: string) => this.deactivatePlugin(name)
+    this.manager['engineActivatePlugin'] = (name: string) => this.activatePlugin(name)
+    this.manager['engineDeactivatePlugin'] = (name: string) => this.deactivatePlugin(name)
+    manager.activatePlugin('manager').then(() => {
       this.isLoaded = true
       // Run callback on `onload` if any
       if (this.managerLoaded) this.managerLoaded()
@@ -133,7 +133,7 @@ export class Engine {
       if (canActivate) {
         await this.manager.toggleActive(target)
       } else {
-        throw new Error(`Cannot call ${method} from ${target}, because ${target} is not activated yet`)
+        throw new Error(`${from.name} cannot call ${method} of ${target}, because ${target} is not activated yet`)
       }
     }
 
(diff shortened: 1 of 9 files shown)


projects/engine/src/engine/engine.ts
```diff
@@ -14,9 +14,9 @@ export class Engine {
   constructor(private manager: BasePluginManager) {
     this.plugins['manager'] = manager
     // Activate the Engine & start listening on activation and deactivation
-    this.activatePlugin('manager').then(() => {
-      this.manager['engineActivatePlugin'] = (name: string) => this.activatePlugin(name)
-      this.manager['engineDeactivatePlugin'] = (name: string) => this.deactivatePlugin(name)
+    this.manager['engineActivatePlugin'] = (name: string) => this.activatePlugin(name)
```
yann300: is this change needed or is it for cleaning the code?

GrandSchtroumpf: It's needed

projects/engine/tests/engine.ts
```diff
@@ -202,14 +202,38 @@ describe('Plugin interaction', () => {
...
+    expect(_currentRequest).toEqual({ from: 'fileManager', path: 'solidity' })
+    expect(updateReq).toHaveBeenCalledTimes(2)
   })
 
   test('Plugin can activate another', async () => {
     await manager.activatePlugin('solidity')
     await solidity.call('manager', 'activatePlugin', 'fileManager')
     const isActive = await manager.isActive('fileManager')
     expect(manager.activatePlugin).toHaveBeenCalledWith('fileManager')
-    expect(manager.canActivate).toHaveBeenCalledWith(solidity.profile, manager.profile)
```
yann300: It looks like `manager.profile` from`expect(manager.canActivate).toHaveBeenCalledWith(solidity.profile, manager.profile)` was not expected cause the plugin that `solidity` try to activate is the `fileManager` . is this directly part of the fix?

yann300: could you point me in the PR where the fix is?

GrandSchtroumpf: The test was wrong and it's one reason why it was hard to detect the bug

GrandSchtroumpf: > could you point me in the PR where the fix is?

I don't understand your question

yann300: if you update the test here (which was before passing I guess), It means that you've added a change in this PR for making sure the test still pass.

GrandSchtroumpf: The change is here : 
https://github.com/ethereum/remix-plugin/pull/171/files/78d3debe2e4c3e2b6bc78e3d4ed3e526fdc865b8#diff-c0e115918fb919bc94b8ec28711c6dd4R17

projects/engine/src/engine/engine.ts
yann300: could you describe why? I see that you are defining the callback before activation but how does that fix the issue (and need a change in the test)?

GrandSchtroumpf: I was doing this.activatePlugin('manager'). Now I'm doing manager.activatePlugin('manager') which does more stuff (including running engine.activatePlugin('manager'))

PR #172, opened 2020-03-13
Implement Secure Ecmascript

GrandSchtroumpf:
Secure Ecmascript (ses) is a library that execute Javascript in a sandboxed environment (called Compartment) without any access to global values (document, console, localstorage, ...).
The host process (remix-ide for example) gives access to variables to the ses compartment: 
```typescript
const c = new Compartment({
  console: console, // compartment can now use the console
  localstorage: {      // compartment will use this implementation of localstorage
    setItem: () => { /* implement a custom localstorage */ }
  }
})
c.evaluate(untrustedCode)
```
- [X] Create client-ses library
- [X] Test for client-ses
- [X] Create a SesPlugin on the engine library
- [ ] Test for SesPlugin (blocked by https://github.com/Agoric/SES-shim/issues/213)
- [ ] Implements PR #176


yann300:
can the untrusted code keep track of their own global variable?

GrandSchtroumpf:
do you have an example I can test ?

yann300:
for instance run:
```
let test = 42
```
and then

```
console.log(test, 'should be 42')
```

GrandSchtroumpf:
Yes you can create a variable and keep track of it. But the `console` won't be accessible if you don't pass it as parameter of the Compartment.

lilvee1365 (commented):
Resolve this branch

PR #178, opened 2020-03-30
Add child process plugin

GrandSchtroumpf:
Create : 
- new project for client-child-process (plugin side)
- new ChildProcessPlugin (engine side)
- add unit tests


projects/client-child-process/jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  name: 'plugin-ws',
```
yann300, line 2: should that be `plugin-child-process`

PR #179, opened 2020-04-04
Only process messages that comes from the current Iframe

yann300:


GrandSchtroumpf:
Hello, can you provide more context why you need to do that ?

yann300:
If the origin of 2 different plugins is the same, we need to check against the `source` (which is unique), please check the issue above, it contains step for reproducing the issue in the IDE.

GrandSchtroumpf:
Ok I think I see. Though this PR should be merged against v0.2.0 as master is reserved for current version in remix ide (not alpha) and will create conflict with next version.

yann300:
moved to https://github.com/ethereum/remix-plugin/pull/181

PR #187, opened 2020-04-08
Fix ethdoc

yann300:
Manage the case where the artifact contains the `results` entry (solidity 0.6)


examples/plugins/ethdoc/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../../../tsconfig.json",
-  "include": ["src", "profile"],
+  "include": ["src", "profile", "../../../webpack.common.js"],
```
GrandSchtroumpf, line 3: I think it's better to take the build output of `@remixproject/plugin-iframe` as this will be the expected entrypoint for iframes plugin. 
So you shouldn' t need any webpack.common.js file.

tsconfig.json
```diff
@@ -14,7 +14,7 @@
       "remix-plugin": ["src/index.ts"],
       "@remixproject/engine": ["./projects/engine/dist"],
       "@remixproject/engine/*": ["./projects/engine/src/*"],
-      "@remixproject/plugin": ["./projects/client/dist"],
+      "@remixproject/plugin": ["./projects/client"],
```
GrandSchtroumpf, line 17: Following the comment above, could you keep this as a reference to "dist" ?

tsconfig.json
```diff
@@ -14,7 +14,7 @@
       "remix-plugin": ["src/index.ts"],
       "@remixproject/engine": ["./projects/engine/dist"],
       "@remixproject/engine/*": ["./projects/engine/src/*"],
-      "@remixproject/plugin": ["./projects/client/dist"],
+      "@remixproject/plugin": ["./projects/client"],
       "@remixproject/plugin/*": ["./projects/client/src/*"],
       "@remixproject/plugin-ws": ["./projects/client-ws"],
       "@remixproject/plugin-iframe": ["./projects/client-iframe"],
```
GrandSchtroumpf, line 20: Follow the comment above this should be `"@remixproject/plugin-iframe": ["./projects/client-iframe/dist"],` to give access to the build file.

webpack.common.js
```diff
@@ -0,0 +1,20 @@
+const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin')
```
GrandSchtroumpf, line 1: Following the comment above we shouldn't have this file.

yann300:
> Do you think we should get ethdoc out of the repo ?

yes @EdsonAlcala is planning to do that

PR #190, opened 2020-04-13
fix ethdoc: return values

yann300:


examples/plugins/ethdoc/webpack.common.js
```diff
@@ -0,0 +1,20 @@
...
+        test: /\.ts?$/,
+        exclude: [/node_modules/],
+        loader: "ts-loader",
+      },
+    ],
+  },
+  resolve: {
+    extensions: ['.ts', '.js'],
+    plugins: [
+      new TsconfigPathsPlugin({ configFile: path.resolve(__dirname, './tsconfig.json') }),
```
GrandSchtroumpf, line 17: I think you want to get the tsconfig.json from the root here instead : `path.resolve(__dirname, '../../../tsconfig.json')`

PR #191, opened 2020-04-13
Client Connector

### LINKED ISSUE #185: [v0.3.0] Create ClientConnector
_implementation of #183_
- [x] Create a conncector interface for the client used to wrap the native connection API:
```typescript
interface ClientConnector {
  /** Send a message to the engine */
  send(message: Partial<Message>): void
  /** Get messae from the engine */
  on((message: Partial<Message>) => void): void
}
```
Implement it for :
- [x] Iframe
- [x] Websocket
- [x] Child Process

GrandSchtroumpf:
fix #185 

Create a client connector & implement it for : 
- Iframe
- websocket
- child process

### DIFF
--- a/projects/client-child-process/connector.ts
+++ b/projects/client-child-process/connector.ts
@@ -10,7 +10,6 @@ export const childProcessConnector: ClientConnector = {
   /** Send a message to the engine */
   send(message: Partial<Message>) {
     process.send(message)
-    this.websocket.send(JSON.stringify(message))
   },
 
   /** Get messae from the engine */
@@ -20,8 +19,8 @@ export const childProcessConnector: ClientConnector = {
 }
 
 /**
- * Connect an Iframe client to a web engine
- * @param client An optional iframe client to connect to the engine.
+ * Connect an Child Process client to a web engine
+ * @param client An optional child process plugin client to connect to the engine.
  *
  * ---------
  * @example
--- a/projects/client-child-process/connector.ts
+++ b/projects/client-child-process/connector.ts
@@ -5,26 +5,17 @@ import { RemixApi } from '../utils/src/api/remix-profile'
 import { Api, ApiMap } from '../utils/src/types/api'
 
 
-export interface WS {
-  send(data: string): void
-  on(type: 'message', cb: (event: string) => any): this
-}
-
-/**
- * This Websocket connector works with the library `ws`
- */
-export class WebsocketConnector implements ClientConnector {
-
-  constructor(private websocket: WS) {}
+export const childProcessConnector: ClientConnector = {
 
   /** Send a message to the engine */
   send(message: Partial<Message>) {
+    process.send(message)
     this.websocket.send(JSON.stringify(message))
-  }
+  },
 
   /** Get messae from the engine */
   on(cb: (message: Partial<Message>) => void) {
-    this.websocket.on('message', (event) => cb(JSON.parse(event)))
+    process.on('message', (event) => cb(event))
   }
 }
 
@@ -35,10 +26,7 @@ export class WebsocketConnector implements ClientConnector {
  * ---------
  * @example
  * ```typescript
- * const wss = new WebSocket.Server({ port: 8080 });
- * wss.on('connection', (ws) => {
- *  const client = createWebsocketClient(ws)
- * })
+ * const client = createChildProcessClient()
  * ```
  * ---------
  * @example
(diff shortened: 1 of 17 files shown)


projects/client-child-process/connector.ts
```diff
@@ -0,0 +1,51 @@
...
+  },
+
+  /** Get messae from the engine */
+  on(cb: (message: Partial<Message>) => void) {
+    process.on('message', (event) => cb(event))
+  }
+}
+
+/**
+ * Connect an Iframe client to a web engine
```
yann300, line 23: you mean `Connect a child process client to a web engine` right?

GrandSchtroumpf: indeed

projects/client-child-process/connector.ts
```diff
@@ -0,0 +1,51 @@
...
+import { RemixApi } from '../utils/src/api/remix-profile'
+import { Api, ApiMap } from '../utils/src/types/api'
+
+
+export const childProcessConnector: ClientConnector = {
+
+  /** Send a message to the engine */
+  send(message: Partial<Message>) {
+    process.send(message)
+    this.websocket.send(JSON.stringify(message))
```
yann300, line 13: Do you really need to `send` two times?

GrandSchtroumpf: ah ! good catch, I was too quick in the copy/paste here.

projects/client-iframe/src/connector.ts
```diff
@@ -0,0 +1,71 @@
...
+      window.parent.postMessage(message, '*')
+    }
+  }
+
+  /** Get messae from the engine */
+  on(cb: (message: Partial<Message>) => void) {
+    window.addEventListener('message', async (event: MessageEvent) => {
+      if (!event.source) throw new Error('No source')
+      // Check that the origin is the right one
+      const devMode = this.client.options.devMode
```
yann300, line 30: If you need to pass `client` only for getting `devMode`, it would be better to either pass only `devMode` or `options`, so we really specify what's needed.
other way would be to create an env variable and use it directly from there.

projects/client-ws/connector.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * This Websocket connector works with the library `ws`
+ */
+export class WebsocketConnector implements ClientConnector {
+
+  constructor(private websocket: WS) {}
+
+  /** Send a message to the engine */
+  send(message: Partial<Message>) {
+    this.websocket.send(JSON.stringify(message))
```
yann300, line 22: should be `websocket.send(JSON.stringify(message))` ?

projects/client-ws/connector.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /** Get messae from the engine */
+  on(cb: (message: Partial<Message>) => void) {
+    this.websocket.on('message', (event) => cb(JSON.parse(event)))
+  }
+}
+
+/**
+ * Connect an Iframe client to a web engine
```
yann300, line 32: `Connect a web socket client to a web engine`

projects/client-ws/connector.ts
```diff
@@ -0,0 +1,66 @@
...
+ * ---------
+ * @example
+ * ```typescript
+ * class MyPlugin extends PluginClient {
+ *  methods = ['hello']
+ *  hello() {
+ *   console.log('Hello World')
+ *  }
+ * }
+ * const wss = new WebSocket.Server({ port: 8080 });
```
yann300, line 52: Where does the WebSocket connection `new WebSocket('ws://localhost:8080', 'echo-protocol')` (the engine trying to connect to the websocket client) actually happen?

projects/client/src/connector.ts
```diff
@@ -0,0 +1,115 @@
...
+ * ```
+ */
+export function connectClient(connector: ClientConnector, client: PluginClient = new PluginClient()) {
+  let isLoaded = false
+
+  connector.on(async ({ action, key, name, payload, id, requestInfo, error }) => {
+    try {
+
+      // If handshake set isLoaded
+      if (isHandshake({ action, key })) {
```
yann300, line 41: perhaps `(isHandshake({ action, key }) && !isLoaded)`

projects/client-ws/connector.ts
GrandSchtroumpf: No, this is the local websocket pass as parameter in the constructor. Using `private` in the parameter of a constructor is a shortcut for making it a property of the class.
I use that so I'm not dependant on the websocket implementation (it's based on the "ws" library, but could be used by another).

projects/client-ws/connector.ts
GrandSchtroumpf: In the server. It's up to the developer to create a websocket connection. I could do it, but I would need to know the client address, and the websocket library. So in the end I think it's better to let the user create his connection and pass it to the plugin client. I'm open to better solutions :)

projects/client/src/connector.ts
GrandSchtroumpf: ok

projects/client-ws/connector.ts
yann300: > In the server

websocket server? not sure how that'd work because the connection is only initiated by remix IDE when the websocket plugin is activated.
The websocket address where to connect to is then given by the `url` property of the profile.

GrandSchtroumpf: Ah sorry, I missunderstood your question. What we call "client" in remix-plugin is the plugin side (which in this case is a websocket server).
The code running inside RemixIDE is the Engine, where we create a new websocket connection when the plugin is activated.

yann300:
could you give a tiny rebase?

GrandSchtroumpf:
On v0.3.0 ? I should first merge on v0.3.0 & then rebase v0.3.0 on master.

PR #195, opened 2020-04-20
Use *.dyn.plugin.remixproject.org for loading ipfs plugin

yann300:


GrandSchtroumpf:
Test are failing

GrandSchtroumpf:
Could this goes for the v0.3.0 release ? Or is it a quick fix for 0.2.0 ?

yann300:
we can put it to `v0.3.0` and backport to `v0.2.0` as soon as it is required

GrandSchtroumpf:
Ok, let's merge it inside master then. I don't really want to cherry-pick stuff in the middle of v0.3.0 for a feature we would want to add inside RemixIDE.
Btw, I think those link should be managed somewhere else for the engine developer to choose where to get the plugin. Maybe some kind of Marketplace plugin

yann300:
> Btw, I think those link should be managed somewhere else for the engine developer to choose where to get the plugin. Maybe some kind of Marketplace plugin

Right, I thought to add an opt in the engine for letting the user define its gateway, but realized that will add too much change and not really in the scope for this PR. But we should definitely not impose this URL.

PR #197, opened 2020-04-24
Increase time out before rejecting with timeout

yann300:


projects/engine/src/plugin/abstract.ts
```diff
@@ -83,7 +83,7 @@ export class Plugin<T extends Api = any, App extends ApiMap = any> implements Pl
           if (this.requestQueue.length !== 0) this.requestQueue[0]()
         }
 
-        const ref = setTimeout(() => { timedout = true, letcontinue() }, 10000)
+        const ref = setTimeout(() => { timedout = true, letcontinue() }, 60000)
```
GrandSchtroumpf, line 86: 1min looks super long. What's your usecase ?

yann300: running the unit test

GrandSchtroumpf: Maybe adding a "enableTimeout = false" property in the engine would be a better fix then I think

yann300: a bit more precision: this is for calling the solidity unittest plugin. it takes often more than 10s for having a result.
> Maybe adding a "enableTimeout = false" property

ok right, but we need timeout anyway , what about adding the opt `timeOut` (in ms) to the engine?

GrandSchtroumpf: Sounds like a good idea

GrandSchtroumpf: Sounds like a good idea

GrandSchtroumpf:
@yann300 Can we close it ? It should be solved by #203

PR #202, opened 2020-04-27
Add documentation

GrandSchtroumpf:
Improve documentation
Update build process to support node & web target
Smal fixes


projects/client/src/connector.ts
```diff
@@ -111,3 +112,22 @@ export function applyApi(client: PluginClient) {
     client[name] = createApi(client, profiles[name])
   }
 }
+
+
+/**
+ * Create & conncect a client with a connector.
```
LianaHus, line 118: typo

projects/engine/doc/api/manager.md
```diff
@@ -0,0 +1,291 @@
...
+    return from.name === 'manager'
+  }
+}
+```
+
+
+> Don't forget to let 'manager' deactivate plugins if you're not using `toggleActivate`.
+
+
+### canCall
```
yann300, line 194: is this `askUserPermission` ?

projects/engine/doc/connector/plugin-connector.md
```diff
@@ -0,0 +1,92 @@
+# Publish a Plugin Connector
```
yann300, line 1: should that be `Develop a Plugin Connector` instead ? where would the connector published?

projects/engine/doc/tutorial/2-plugin-communication.md
```diff
@@ -0,0 +1,134 @@
+## Plugins Communication
+
+### Methods
+
+Each plugin can call methods exposed by other plugin. Let's see how to expose a method from one plugin and call it from another.
+
+1. Create Plugin that exposes a method
```
yann300, line 7: sounds that this can also be useful for plugin developer?

projects/engine/doc/api/manager.md
GrandSchtroumpf: In version 0.1.0 yes. Since version 0.2.0 we separate askUserPermission into tow method: 
- `canActivate`: if plugin can ativate another
- `canCall`: If a plugin can call a method from another.

projects/engine/doc/connector/plugin-connector.md
GrandSchtroumpf: On npm. I'll add `Develop & Publish`.

projects/engine/doc/tutorial/2-plugin-communication.md
GrandSchtroumpf: Not really. There is a big problem of naming with the remix-plugin right now because : 
- `Plugin` is the code that runs in the engine
- `Client` is the code that run in the plugin (build by the plugin developer)

In this case we are talking about the `Plugin` in the engine.

projects/engine/doc/tutorial/2-plugin-communication.md
```diff
@@ -0,0 +1,134 @@
...
+You can extends the `Plugin` class to create your own plugin. The list of exposed methods are defined in the field `methods` of the profile: 
+```typescript
+class FirstPlugin extends Plugin {
```
yann300, line 10: I mean that ^

> sounds that this can also be useful for plugin developer?

> Not really. There is a big problem of naming with the remix-plugin right now because :
> Plugin is the code that runs in the engine
> Client is the code that run in the plugin (build by the plugin developer)
> In this case we are talking about the Plugin in the engine.

PR #203, opened 2020-05-04
Plugin Option System

GrandSchtroumpf:
Create an option system on plugins : 
- Each plugin type can have specific option (extranal plugins have transformUrl for example).
- Engine has `addPluginOption` to apply global options to each plugin
- A plugin can override global options with `setOptions` (can be used inside `onRegistration` for example).

Here is the life cycle of plugin options
1. option is set in the plugin constructor
2. engine update options with `addPluginOption`
3. plugin can update options within `onRegistration`

### DIFF
--- a/projects/engine/src/plugin/external.ts
+++ b/projects/engine/src/plugin/external.ts
@@ -1,6 +1,6 @@
 import { ExternalProfile, Profile } from '../../../utils/src/types/profile'
 import { Message } from '../../../utils/src/types/message'
-import { Plugin } from './abstract'
+import { Plugin, PluginOptions } from './abstract'
 
 /** List of available gateways for decentralised storage */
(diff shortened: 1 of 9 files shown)


projects/engine/src/plugin/external.ts
```diff
@@ -31,6 +35,11 @@ export abstract class ExternalPlugin extends Plugin {
     super.deactivate()
   }
 
+  /** Set options for an external plugin */
+  setOptions(options: Partial<ExternalPluginOptions> = {}) {
```
yann300, line 39: is this needed to overrude this? `setOptions` will anyway be available

GrandSchtroumpf: For the type in the params

PR #205, opened 2020-05-11
First PR on vscode connector

GrandSchtroumpf:
Create connector for webview on engine

### DIFF
--- a/projects/engine-vscode/jest.config.js
+++ b/projects/engine-vscode/jest.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  name: 'engine-web',
+  preset: '../../jest.config.js',
+  coverageDirectory: '../../coverage/engine-web'
+};
--- a/projects/engine-vscode/src/webview.ts
+++ b/projects/engine-vscode/src/webview.ts
@@ -1,0 +1,83 @@
+import { PluginConnector, Profile, Message, PluginConnectorOptions, ExternalProfile } from "@remixproject/engine"
+import { ExtensionContext, ViewColumn, WebviewPanel, window, Uri } from 'vscode'
+import { join } from 'path'
+import { promises as fs, watch } from 'fs'
+
+interface WebviewOptions extends PluginConnectorOptions {
+  /** Extension Path */
+  context: ExtensionContext
+  column: ViewColumn
+}
+
+export class WebviewPlugin extends PluginConnector {
+  panel?: WebviewPanel
+  options: WebviewOptions
+
+  constructor(profile: Profile & ExternalProfile, options: WebviewOptions) {
+    super(profile)
+    this.options = options
+  }
+
+  protected send(message: Partial<Message>): void {
+    if (this.panel) {
+      this.panel.webview.postMessage(message)
+    }
+  }
+
+  protected connect(url: string): void {
+    const { extensionPath } = this.options.context
+    this.panel = createWebview(extensionPath, this.name, this.options.column)
+    this.panel.webview.onDidReceiveMessage(msg => this.getMessage(msg))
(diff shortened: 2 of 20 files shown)


projects/engine-vscode/jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
yann300, line 1: `engine-vscode` ?

projects/engine/src/connector.ts
```diff
@@ -1,25 +1,29 @@
...
 /** List of available gateways for decentralised storage */
 export const defaultGateways = {
   'ipfs://': (url, name) => `https://${name}.dyn.plugin.remixproject.org/ipfs/${url.replace('ipfs://', '')}`,
-  'swarm://': (url, _) => `https://swarm-gateways.net/bzz-raw://${url.replace('swarm://', '')}`
+  'swarm://': (url, name) => `https://swarm-gateways.net/bzz-raw://${url.replace('swarm://', '')}`
 }
 
 /** Transform the URL to use a gateway if decentralised storage is specified */
-export function transformUrl(url: string, name: string) {
+export function transformUrl({ url, name }: Profile & ExternalProfile) {
```
yann300, line 12: I "think" this is part of a previous PR.

GrandSchtroumpf: It is, but the "connector" file doesn't exist on master

projects/engine-vscode/src/webview.ts
```diff
@@ -0,0 +1,140 @@
...
+      this.panel.webview.postMessage(message)
+    }
+  }
+
+  protected connect(url: string): void {
+    if (this.options.context) {
+      const { extensionPath } = this.options.context
+      this.panel = createWebview(this.profile, url, extensionPath, this.options.column)
+      this.panel.webview.onDidReceiveMessage(msg => this.getMessage(msg))
+      this.options.context.subscriptions.push(this.panel)
```
yann300, line 38: can you give more context on what is `subscriptions` ?

GrandSchtroumpf: It's a vscode stuff to keep track of the panel.

projects/engine-vscode/src/webview.ts
```diff
@@ -0,0 +1,140 @@
...
+    return `${prefix}="${path}"`
+  }
+
+  const html = await fetch(index)
+  webview.html = html.replace(matchLinks, toRemoteUrl)
+}
+
+
+/** Get code from local source */
+async function setLocalHtml(webview: Webview, baseUrl: string) {
```
yann300, line 120: I guess local resources need to be packaged in the extension?

GrandSchtroumpf: What do you mean ?

yann300: if a webview want to load local resources, the local resources need to be packaged in the extension.
will try to find a more concrete example.

GrandSchtroumpf: Yes. Let's take an exemple : 
```
src/
  extension/ --> folder with extension code, engine, ..
  webview/  --> folder with webview code
dist/
  extension/
    webview/ --> build result of the webview code
    main.js     --> entrypoint of the extension
    package.json
```
This is because when you're going to publish your extension on vscode marketplace you'll publish the extension folder, so the webview must be inside.

yann300: ok thanks, 
If one try to load a resources (http or internal) from the code loaded in the webview ( like with a script tag), I've seen there is some limitation for external resources, but it seem one can load internal resources. But what is exactly an internal resources? is this something that need to be package within the extension?

PR #206, opened 2020-05-18
Finish implementation of engine vscode

GrandSchtroumpf:
Now we can run an engine on vscode that supports webview.
Note: the client side connector is not done yet.


GrandSchtroumpf:
Wrong branch to merge on

PR #207, opened 2020-05-18
Engine vscode

GrandSchtroumpf:
Now we can run an engine on vscode that supports webview.
Note: the client side connector is not done yet.


projects/engine-node/tests/child-process.ts
```diff
@@ -81,7 +80,7 @@ describe('ChildProcess plugin', () => {
     try {
       await plugin.activate();
       (plugin.process as any).connected = false
-      plugin['postMessage']({ name: 'child-process' })
+      plugin['send']({ name: 'child-process' })
```
yann300, line 83: what's the reasoning behind this renaming?

projects/engine-vscode/src/webview.ts
```diff
@@ -1,14 +1,18 @@
...
-import { promises as fs } from 'fs'
+import { promises as fs, watch } from 'fs'
 import { get } from 'https'
 import { parse as parseUrl } from 'url'
 
 interface WebviewOptions extends PluginConnectorOptions {
   /** Extension Path */
   context: ExtensionContext
   column?: ViewColumn
+  devMode?: boolean
```
yann300, line 15: should `devMode`  be a property of `PluginConnectorOptions` ?

projects/engine-vscode/src/webview.ts
```diff
@@ -1,14 +1,18 @@
...
 import { join, parse as parsePath } from 'path'
-import { promises as fs } from 'fs'
+import { promises as fs, watch } from 'fs'
 import { get } from 'https'
 import { parse as parseUrl } from 'url'
 
 interface WebviewOptions extends PluginConnectorOptions {
   /** Extension Path */
   context: ExtensionContext
   column?: ViewColumn
```
yann300, line 14: although all the other properties are self explanatory, could you add a comment for that one?

projects/engine-node/tests/child-process.ts
GrandSchtroumpf: It was either `postMessage`, the iframe style, or `send`, the websocket style. I choose the shortest.

projects/engine-vscode/src/webview.ts
GrandSchtroumpf: Could be yes.

projects/engine-vscode/src/webview.ts
GrandSchtroumpf: Sure. For a vscode developer it's quite obvious. When you divide your editor in several column it's the index in which you want to add your plugin. A little bit like the location in remix-ide

projects/engine-web/tests/iframe.ts
```diff
@@ -115,60 +113,12 @@ describe('Iframe Plugin', () => {
...
-      data: { id: 0, action: 'request', key: 'method', payload: ['params'], name: 'manager' }
-    }
-    iframe['getEvent'](event as any)
-    expect(iframe.call).toHaveBeenCalledWith('manager', 'method', 'params')
-    const response = { id: 0, action: 'response', key: 'method', payload: [true], name: 'manager', error: undefined }
-    setTimeout(() => expect(spy).toHaveBeenCalledWith(response), 10)  // Wait for next tick
-  })
-
-  // Get Message: listen
-  test('getEvent with listen should run "on"', () => {
```
yann300, line 133: are these tests not used anymore?

GrandSchtroumpf: I move them into `engine-web`

projects/engine/src/connector.ts
```diff
@@ -77,11 +78,14 @@ export abstract class PluginConnector extends Plugin {
   protected async handshake() {
     if (!this.loaded) {
       const methods: string[] = await this.callPluginMethod('handshake', [this.profile.name])
-      this.call('manager', 'updateProfile', this.profile)
-      this.loaded = true
       if (methods) {
         this.profile.methods = methods
       }
+      this.call('manager', 'updateProfile', this.profile)
+      this.loaded = true
```
yann300, line 85: If the `loaded` is set at the end (especially after the `await` - line 80), it is probable that `this.callPluginMethod('handshake',  ... ` can be called several times.

yann300: e.g if 2 handshake are triggered before the execution reach the `this.loaded = true`

GrandSchtroumpf: Ok, but if handshake fails you don't want to set it to loaded...
I don't see the risk of having two handshakes except updating twice the profile methods.

PR #208, opened 2020-05-21
Redirect Filemanager API

ioedeveloper:
Implements redirect from old api to new api https://github.com/ethereum/remix-ide/pull/2733/commits/14605cc80225e52b89330ae223463055b8aab293

### DIFF
--- a/projects/engine/src/engine/engine.ts
+++ b/projects/engine/src/engine/engine.ts
@@ -140,13 +140,18 @@ export class Engine {
     }
 
     // Check if method is exposed
-    if (!to.methods.includes(method)) {
+    const request = { from: caller, path }
+
+    if (to.redirect) {
+      const redirectMethods = Object.keys(to.redirect)
+
+      if (redirectMethods.includes(method)) return this.plugins[target]['addRequest'](request, to.redirect[method], payload)
+    } else if (!to.methods.includes(method)) {
       const notExposedMsg = `Cannot call method "${method}" of "${target}" from "${caller}", because "${method}" is not exposed.`
       const exposedMethodsMsg = `Here is the list of exposed methods: ${to.methods.map(m => `"${m}"`).join(',')}`
       throw new Error(`${notExposedMsg} ${exposedMethodsMsg}`)
     }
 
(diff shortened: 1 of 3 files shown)


GrandSchtroumpf:
I don't see why it should be integrated inside remix-engine. This logic should be managed in remix-ide : 
```typescript
constructor() {
  this.getCurrentFile = this.file
  this.getFile = this.readFile
}
```

yann300:
I see that redirection are defined in remix-ide
```redirect: {
    getCurrentFile: 'file',
    getFile: 'readFile',
    getFolder: 'readdir',
    setFile: 'writeFile',
    switchFile: 'open'
  },```

we should put this in remix-plugin, other platform should not need to take care of this.

GrandSchtroumpf:
There are no other platform running remix engine yet. And only the one running inside a the web would benefit from it as the other (node & vscode) would require the current plugin to update to a higher version anyway. I hope they would update their code by then.
I understand the idea of keeping the engine retrocompatible with every version of the any engine that existed. But I think it's a very difficult task to tacle. And I would vote against doing that as it will make changes more difficult, and I think its too early in the life of the engine to burden ourselve with that. Maybe when we hit 1.0.0 with a better vision on how developers are using this libraries we can define a better multiplatform strategy.

yann300:
>  would require the current plugin to update to a higher version anyway

You would suggest to propose them the following: 
 "if you want to be published in vscode, please update your remix-plugin version to latest one and update the following API:"
  "fileManager: getFile => readFile, etc ..."

I am not necessarily against that but we should really make the necessary to not force them to push complex changes, and to document it  as much as possible.
This won't require changes in the input paramaters right?

GrandSchtroumpf:
For sure we should document the migration process. I don't know exactly ohw to make plugin cross platform, I still need to work on that. The goal is that they don't have to change their code (just new npm install + some config).

The update of API is another subject. It's totally up to the IDE to decide the migration strategy. For me the easier is to continue to support previous endpoints with same arguments and redirect it to the new API. But I would recommend to force yourself to keep exactly the same arguments for the new API if you need a different one.

PR #209, opened 2020-05-25
Add client vscode

GrandSchtroumpf:
Improve webview to support absolute path
Create `CommandPlugin`
Create client vscode

### DIFF
--- a/projects/engine-vscode/src/command.ts
+++ b/projects/engine-vscode/src/command.ts
@@ -9,6 +9,9 @@ interface CommandOptions extends PluginOptions {
 }
 
 // WIP
+/**
+ * Connect methods of the plugins with a command depending on the transformCmd function pass as option
+ */
 export class CommandPlugin extends Plugin {
   subscriptions: Disposable[] = []
   options: CommandOptions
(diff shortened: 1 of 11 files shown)


projects/engine-vscode/src/command.ts
```diff
@@ -0,0 +1,35 @@
+import { Plugin, Profile, PluginOptions } from "@remixproject/engine"
+import { Disposable, commands } from "vscode"
+
+export const transformCmd = (name: string, method: string) => `${name}.${method}`
+
+interface CommandOptions extends PluginOptions {
+  transformCmd: (name: string, method: string) => string
+}
+
+export class CommandPlugin extends Plugin {
```
yann300, line 10: Could you give some detail here about the place of `command` in the VScode stack?

GrandSchtroumpf: It's a record of every command you can run in vscode. When you do `ctrl+p` you can look for all the command available.
Omkara needs that to connect command on activation / deactivation of plugins. I'll add a comment "WIP".

yann300: Thanks for the precision, can you add a class comment too?

PR #217, opened 2020-06-22
Unique connector webview / Iframe

GrandSchtroumpf:


projects/client-vscode/src/webview.ts
```diff
@@ -1,51 +1,76 @@
...
+  PluginOptions,
+  checkOrigin
+} from '@remixproject/plugin'
 
 declare const acquireVsCodeApi: any
-
 /**
  * This Webview connector
  */
 export class WebviewConnector implements ClientConnector {
```
yann300, line 19: shouldn't this be in a `connector.ts` (like for iframe.ts) or is this fine to keep it here?

projects/client-vscode/src/webview.ts
```diff
@@ -1,51 +1,76 @@
...
 
+
   /** Send a message to the engine */
   send(message: Partial<Message>) {
-    this.postMessage(message)
+    if (this.isVscode) {
+      this.source.postMessage(message)
+    } else if (this.origin || isHandshake(message)) {
+      const origin = this.origin || '*'
+      this.source.postMessage(message, origin)
```
yann300, line 36: don't need to use `window.parent.` anymore ? like for https://github.com/ethereum/remix-plugin/blob/v0.3.0/projects/client-iframe/src/connector.ts#L21

webpack.config.js
```diff
@@ -133,6 +133,9 @@ const wsClient = {
 // Iframe client
 const iframeClient = {
   ...config('client-iframe'),
+  externals: {
+    '@remixproject/plugin': 'commonjs2 @remixproject/plugin',
```
yann300, line 137: is this file packaged `commonjs2` ?

projects/client-vscode/src/webview.ts
GrandSchtroumpf: It's the default source when parent s not vscode. : line 26
`this.source = this.isVscode ? acquireVsCodeApi() : window.parent`

projects/client-vscode/src/webview.ts
GrandSchtroumpf: I think `connector.ts` in iframe should be renamed. But I'm still figuring out how to organize this repo. I' happy to update, but I would prefer to se how it comes out.

PR #221, opened 2020-07-09
Throw error back to the engine

yann300:


projects/@remixproject
```diff
@@ -0,0 +1 @@
+node_modules/@remixproject
```
GrandSchtroumpf, line 1: What is it for ?

projects/projects
```diff
@@ -0,0 +1 @@
+/home/yann/Remix/remix-plugin/remix-plugin/projects
```
GrandSchtroumpf, line 1: What is it for ?

projects/@remixproject
yann300: removing it

PR #235, opened 2020-08-31
Add native vscode plugins

0mkara:


package.json
```diff
@@ -51,5 +51,8 @@
     "ts-node": "~7.0.0",
     "tslint": "~6.0.0",
     "typescript": "3.9.7"
+  },
+  "dependencies": {
```
GrandSchtroumpf, line 55: This should be part of the `packages/engine/vscode/package.json` instead except if you added unit tests that require to have the dependencies in the global project.

packages/engine/vscode/src/lib/compile_worker.ts
```diff
@@ -0,0 +1,118 @@
+import * as solc from "solc";
+import * as path from "path";
+import * as fs from "fs";
+import axios from "axios";
```
GrandSchtroumpf, line 4: If possible don't add unnecessary external dependencies. You can reuse the `fetch` function in webview.ts of the same folder. It looks like that : 
```typescript
/** Fetch remote ressource with http */
function fetch(url: string): Promise<string> {
  return new Promise((resolve, reject) => {
    get(url, res => {
      let text = ''
      res.on('data', data => text += data)
      res.on('end', _ => resolve(text))
      res.on('error', err => reject(err))
    })
  })
}
```

packages/engine/vscode/src/lib/compile_worker.ts
```diff
@@ -0,0 +1,118 @@
...
+import { RemixURLResolver } from "remix-url-resolver";
+
+
+function handleLocal(pathString: string, filePath: any) {
+	// if no relative/absolute path given then search in node_modules folder
```
GrandSchtroumpf, line 9: please use 2spaces for the formating.

packages/engine/vscode/src/lib/compile_worker.ts
```diff
@@ -0,0 +1,118 @@
...
+	if (pathString && pathString.indexOf(".") !== 0 && pathString.indexOf("/") !== 0) {
```
GrandSchtroumpf, line 10: Can you use `path.isAbsolute()` instead ?

packages/engine/vscode/src/lib/compile_worker.ts
```diff
@@ -0,0 +1,118 @@
...
+		console.error("Error: Node Modules Import is not implemented yet!");
+		// return handleNodeModulesImport(pathString, filePath, pathString)
+		return;
+	} else {
+		try {
+			const o = { encoding: "UTF-8" };
+			// hack for compiler imports to work (do not change)
+			const p = pathString ? path.resolve(pathString, filePath) : path.resolve(pathString, filePath);
```
GrandSchtroumpf, line 18: prefere `join` to `resolve` as it takes the OS syntax into account

GrandSchtroumpf: Lools like you're doing the same thing in both case.

packages/engine/vscode/src/lib/compile_worker.ts
```diff
@@ -0,0 +1,118 @@
...
+			const content = fs.readFileSync(p, o);
```
GrandSchtroumpf, line 19: Prefer asynchronous reading: `return fs.promise.readFile(p, o)`

packages/engine/vscode/src/lib/compile_worker.ts
```diff
@@ -0,0 +1,118 @@
...
+import * as path from "path";
+import * as fs from "fs";
+import axios from "axios";
+import { RemixURLResolver } from "remix-url-resolver";
+
+
+function handleLocal(pathString: string, filePath: any) {
+	// if no relative/absolute path given then search in node_modules folder
+	if (pathString && pathString.indexOf(".") !== 0 && pathString.indexOf("/") !== 0) {
+		console.error("Error: Node Modules Import is not implemented yet!");
```
GrandSchtroumpf, line 11: Maybe throw an error instead ?

PR #236, opened 2020-08-31
Add filemanager plugin

0mkara:
- [x] vscode specific native filemanger plugin

### DIFF
--- a/packages/engine/vscode/src/lib/filemanager.ts
+++ b/packages/engine/vscode/src/lib/filemanager.ts
@@ -32,6 +32,7 @@ const profile = {
  * it should be possible to install https://github.com/ethereum/remix-plugins-directory and
  * import `NextFileSystemApi` interface and implement `FileManagerPlugin` following that interface
  * `export default class FileManagerPlugin extends CommandPlugin implements NextFileSystemApi`
+ * 
  */
 
 export default class FileManagerPlugin extends CommandPlugin {
--- a/packages/engine/vscode/src/lib/filemanager.ts
+++ b/packages/engine/vscode/src/lib/filemanager.ts
@@ -49,7 +49,7 @@ export default class FileManagerPlugin extends CommandPlugin {
     const uri = Uri.parse(path);
     const encoder = new TextEncoder();
     const uint8Array = encoder.encode(data);
-    return workspace.fs.writeFile(uri, Uint8Array.from(uint8Array)).then(() => { });
+    return workspace.fs.writeFile(uri, Uint8Array.from(uint8Array));
   }
   /** Return the content of a specific file */
   readFile(path: string): Thenable<String> {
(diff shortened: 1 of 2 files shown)


packages/engine/vscode/src/lib/filemanager.ts
```diff
@@ -0,0 +1,150 @@
...
+ * `export default class FileManagerPlugin extends CommandPlugin implements NextFileSystemApi`
+ */
+
+export default class FileManagerPlugin extends CommandPlugin {
+  constructor() {
+    super(profile);
+  }
+  /** Open the content of the file in the context (eg: Editor) */
+  open(path: string): Promise<boolean> {
+    return new Promise((resolve, reject) => {
```
GrandSchtroumpf, line 44: You don't need to create a promise here : 
```typescript
open(path: string): Promise<boolean> {
  const uri = Uri.file(path);
  commands.executeCommand('vscode.open', uri).then(() => true);
}
```

GrandSchtroumpf: This apply for all below

packages/engine/vscode/src/lib/filemanager.ts
```diff
@@ -0,0 +1,150 @@
...
+        .then(() => {
+          resolve(true);
+        })
+    })
+  }
+  /** Set the content of a specific file */
+  writeFile(path: string, data: string): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const turi = Uri.parse(path);
+      const encoder = new TextEncoder();
```
GrandSchtroumpf, line 56: Where does the TextEncoder comes from ?

packages/engine/vscode/src/lib/filemanager.ts
```diff
@@ -0,0 +1,150 @@
...
+      const uint8Array = encoder.encode(data);
+      workspace.fs.writeFile(turi, Uint8Array.from(uint8Array)).then(() => {
+        resolve();
+      });
+    });
+  }
+  /** Return the content of a specific file */
+  readFile(path: string): Promise<String> {
```
GrandSchtroumpf, line 64: Just a little improvement : 
```typescript
readFile(path: string): Promise<String> {
  const uri = Uri.parse(path);
  return workspace.fs.readFile(uri).then(content => Buffer.from(content).toString("utf-8"))
}
```

packages/engine/vscode/src/lib/filemanager.ts
```diff
@@ -0,0 +1,150 @@
...
+      const duri = Uri.parse(path);
+      workspace.fs.readDirectory(duri).then((data) => {
+        resolve(data);
+      });
+    });
+  }
+  // ------------------------------------------
+  // Legacy API. To be removed.
+  // ------------------------------------------
+  async getFolder(path: string): Promise<Folder> {
```
GrandSchtroumpf, line 116: It should only be a redirect to `readdir` : 
```typescript
getFolder = readdir;
getFile = readFile;
setFile = writeFile;
switchFile = open;
getCurrentFile() {
  return window.activeTextEditor?.document?.fileName;
}
```

GrandSchtroumpf (changes requested):

packages/engine/vscode/src/lib/filemanager.ts
0mkara: From typescript.

ref: https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder

packages/engine/vscode/src/lib/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface Folder {
```
GrandSchtroumpf, line 1: Normally you don't need that anymore

GrandSchtroumpf (approved):

GrandSchtroumpf (approved):

PR #238, opened 2020-09-07
add editor plugin code

0mkara:


packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,75 @@
...
+    return editor.setDecorations(this.decoration, [newDecoration]);
+  }
+  discardHighlight(): void {
+    return this.decoration.dispose();
+  }
+  discardHighlightAt(line: number, filePath: string): void {}
+  addAnnotation(annotation: Annotation): void {
+    // This function should append to existing map
+    // Ref: https://code.visualstudio.com/api/language-extensions/programmatic-language-features#provide-diagnostics
+    const fileUri = window.activeTextEditor ? window.activeTextEditor.document.uri : undefined; // TODO: we might want to supply path to addAnnotation function
```
GrandSchtroumpf, line 50: Yes : 
```typescript
private getEditor(filePath?: string) {
  return path
    ? editors.find(editor => editor.document.uri.path === Uri.parse(filePath).path)
    : window.activeTextEditor
}
addAnnotation(annotation: Annotation, filePath?: string) {
  const editor = this.getEditor(filePath);
  if (editor) {
    ...
  }
}
```

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,75 @@
...
+    let diagnosticMap: Map<string, Diagnostic[]> = new Map();
+    let canonicalFile = fileUri.toString();
```
GrandSchtroumpf, line 52: Why no `uri.fsPath` ?

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,75 @@
...
+  }
+  discardHighlight(): void {
+    return this.decoration.dispose();
+  }
+  discardHighlightAt(line: number, filePath: string): void {}
+  addAnnotation(annotation: Annotation): void {
+    // This function should append to existing map
+    // Ref: https://code.visualstudio.com/api/language-extensions/programmatic-language-features#provide-diagnostics
+    const fileUri = window.activeTextEditor ? window.activeTextEditor.document.uri : undefined; // TODO: we might want to supply path to addAnnotation function
+    let diagnosticMap: Map<string, Diagnostic[]> = new Map();
```
GrandSchtroumpf, line 51: use const (same for all below)

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,75 @@
...
+    let canonicalFile = fileUri.toString();
+    let range = new Range(annotation.row - 1, annotation.column, annotation.row - 1, annotation.column);
+    let diagnostics = diagnosticMap.get(canonicalFile);
+    if (!diagnostics) { diagnostics = []; }
+    let svrt: DiagnosticSeverity = 0;
```
GrandSchtroumpf, line 56: Better to have a mapping : 
```typescript
const diagnosticSeverity: Record<string, DiagnosticSeverity> = {
  'error': DiagnosticSeverity.Error,
  'warning': DiagnosticSeverity.Warning,
  ...
}
...
const severity = diagnosticSeverity[annotation.type];
```

GrandSchtroumpf (changes requested):

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,93 @@
-// Look at this : https://microsoft.github.io/language-server-protocol/specifications/specification-current/
\ No newline at end of file
+import { CommandPlugin, PluginOptions } from "@remixproject/engine-vscode";
+import { Annotation, HighlightPosition } from '@remixproject/plugin-api';
+import { window, Range, TextEditorDecorationType, Position, languages, DiagnosticCollection, Diagnostic, Uri, DiagnosticSeverity, TextEditor } from "vscode";
+
+const profile = {
```
GrandSchtroumpf, line 5: You should be able to import it from @remixproject/plugin-api too. Check how I did it in filemanager

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,93 @@
...
+  permission: true,
+  location: "sidePanel",
+  documentation: "https://remix-ide.readthedocs.io/en/latest/solidity_editor.html",
+  version: "0.0.1",
+  methods: ["highlight", "discardHighlight", "addAnnotation", "clearAnnotations"],
+};
+interface EditorOptions extends PluginOptions {
+  language: string;
+}
+export default class EditorPlugin extends CommandPlugin {
```
GrandSchtroumpf, line 19: Don't use default

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,93 @@
...
+    const diagnosticSeverity: Record<string, DiagnosticSeverity> = {
+      'error': DiagnosticSeverity.Error,
+      'warning': DiagnosticSeverity.Warning,
+      'information': DiagnosticSeverity.Information
+    };
+    const severity = diagnosticSeverity[annotation.type];
+    diagnostics.push(new Diagnostic(range, annotation.text, severity));
+    diagnosticMap.set(canonicalFile, diagnostics);
+    diagnosticMap.forEach((diags, file) => {
+      this.diagnosticCollection.set(Uri.parse(file), diags);
```
GrandSchtroumpf, line 86: Are you sure Uri.parse is working ? From my experience you need to use Uri.file instead.

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,93 @@
...
+    });
+    return;
```
GrandSchtroumpf, line 88: No need to add return.

GrandSchtroumpf (changes requested):

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,84 @@
...
+  }
+  setOptions(options: EditorOptions) {
+    super.setOptions(options);
+  }
+  onActivation() {
+    this.decoration = window.createTextEditorDecorationType({
+      backgroundColor: 'editor.lineHighlightBackground',
+      isWholeLine: true,
+    });
+    // @ts-ignore
```
GrandSchtroumpf, line 25: Why do you need to ignore here ?

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,84 @@
...
+      const start: Position = new Position(position.start.line, position.start.column);
+      const end: Position = new Position(position.end.line, position.end.column);
+      const newDecoration = { range: new Range(start, end) };
+      if (hexColor) {
+        this.decoration = window.createTextEditorDecorationType({
+          backgroundColor: hexColor,
+          isWholeLine: true,
+        });
+      }
+      return editor.setDecorations(this.decoration, [newDecoration]);
```
GrandSchtroumpf, line 45: If `setDecorations` returns void, then don't return it at all. Else update the output type of the method.

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,84 @@
...
+    } else {
+      throw new Error(`Could not find file ${filePath}`);
+    }
+  }
+  async discardHighlight(): Promise<void> {
+    return this.decoration.dispose();
+  }
+  async discardHighlightAt(line: number, filePath: string): Promise<void> {
```
GrandSchtroumpf, line 53: Remove the params and add a comment : 
```typescript
/**
 * Alisas of  discardHighlight
 * Required to match the standard interface of editor
 */
```

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,84 @@
...
+    return this.decoration.dispose();
+  }
+  private getEditor(filePath?: string): TextEditor {
+    const editors = window.visibleTextEditors;
+    return filePath ? editors.find(editor => editor.document.uri.path === Uri.parse(filePath).path) : window.activeTextEditor
```
GrandSchtroumpf, line 58: Is Uri.parse working on Windows ?

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,84 @@
...
+      throw new Error(`Could not find file ${filePath}`);
+    }
+  }
+  async discardHighlight(): Promise<void> {
+    return this.decoration.dispose();
+  }
+  async discardHighlightAt(line: number, filePath: string): Promise<void> {
+    return this.decoration.dispose();
+  }
+  private getEditor(filePath?: string): TextEditor {
```
GrandSchtroumpf, line 56: This method doesn't need to be part of the class (doesn't use `this` and is not exposed). Put it outside of the class declaration as a standalone function.

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,84 @@
...
+    const editors = window.visibleTextEditors;
+    return filePath ? editors.find(editor => editor.document.uri.path === Uri.parse(filePath).path) : window.activeTextEditor
+  }
+  async addAnnotation(annotation: Annotation, filePath?: string): Promise<void> {
+    // This function should append to existing map
+    // Ref: https://code.visualstudio.com/api/language-extensions/programmatic-language-features#provide-diagnostics
+    // const fileUri = window.activeTextEditor ? window.activeTextEditor.document.uri : undefined; // TODO: we might want to supply path to addAnnotation function
+    const editor = this.getEditor(filePath);
+    const canonicalFile: string = editor.document.uri.fsPath;
+    const diagnosticMap: Map<string, Diagnostic[]> = new Map();
```
GrandSchtroumpf, line 66: I don't see the point of using a Map structure if you're using a string as key. `Map` structure is useful when you want to use a reference (object for example) as a key. 
Using a Javascript Object structure (Record) would be lighter I think.

packages/engine/vscode/src/lib/editor.ts
```diff
@@ -1 +1,84 @@
...
+    const range = new Range(annotation.row - 1, annotation.column, annotation.row - 1, annotation.column);
+    const diagnostics = diagnosticMap.get(canonicalFile) ? diagnosticMap.get(canonicalFile) : [];
```
GrandSchtroumpf, line 68: How is it possible that you get a diagnostic from the `diagnosticMap` ? You just initialize the map two lines above and you don't mutate it in-between.
The way I see it you should have : 
```typescript
const editor = this.getEditor(filePath);
const path = editor.document.uri.fsPath;
const range = new Range(annotation.row - 1, annotation.column, annotation.row - 1, annotation.column);
const severity = diagnosticSeverity[annotation.type];
const diagnostic = new Diagnostic(range, annotation.text, severity);
this.diagnosticCollection?.set(Uri.file(path), [ diagnostic ]);
```

GrandSchtroumpf (changes requested):

packages/engine/vscode/src/lib/editor.ts
0mkara: I am unable to check that.

README.md
```diff
@@ -73,6 +73,11 @@ This uses nx's affected:build to only update what has been changes since last bu
 npm run build
 ```
 
+## Build `engine-vscode`
```
GrandSchtroumpf, line 76: Better: 
```typescript
## Build a specific project
npx nx build ${projectName} --with-deps

Example for engine-vscode :
npx nx build engine-vscode --with-deps
```

```

File: packages/engine/vscode/src/lib/editor.ts [hunk 17]
```diff
@@ -1 +1,87 @@
...
+    const canonicalFile: string = editor.document.uri.fsPath;
+    const diagnostics: Diagnostic[] = [];
+    const range = new Range(annotation.row - 1, annotation.column, annotation.row - 1, annotation.column);
+    const diagnosticSeverity: Record<string, DiagnosticSeverity> = {
+      'error': DiagnosticSeverity.Error,
+      'warning': DiagnosticSeverity.Warning,
+      'information': DiagnosticSeverity.Information
+    };
+    const severity = diagnosticSeverity[annotation.type];
+    const diagnostic: Diagnostic = new Diagnostic(range, annotation.text, severity);
```
GrandSchtroumpf, line 80: You shouldn't need to force type it to `Diagnostic`. Let typescript do the job for you when possible ;)

GrandSchtroumpf (approved):

PR #239, opened 2020-09-07
Create terminal plugin for vscode

GrandSchtroumpf:


packages/engine/vscode/src/lib/terminal.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+    return this.active;
+  }
+  
+  private getOutput(name?: string) {
+    name = name ?? this.activeOutput ?? 'plugin-engine';
+    return this.outputs[name] ?? window.createOutputChannel(name);
+  }
+
+  /** Open specific terminal (doesn't work with output) */
```
yann300, line 55: `doesn't work with output` what does that exactly means?

yann300 (approved):

packages/engine/vscode/src/lib/terminal.ts
GrandSchtroumpf: In vscode outputs (stdout) & terminal (stdin) are two different tabs. When you run `open` it can open a terminal but not an output. I could add a second parameter to specify if you want to open a terminal or output. But it's very specific to vscode and if we want to use this API in remix-ide for example, this params wouldn't make sense.
I would prefer having some feedback from plugin developers before going to deep into this implementation.

PR #243, opened 2020-09-14
Documentation and fixes

GrandSchtroumpf:
Improve documentation for vscode engine
Add some fixes on manager & webview

### DIFF
--- a/packages/engine/vscode/src/lib/filemanager.ts
+++ b/packages/engine/vscode/src/lib/filemanager.ts
@@ -1,7 +1,16 @@
 import { filSystemProfile, IFileSystem } from '@remixproject/plugin-api'
 import { MethodApi } from '@remixproject/plugin-utils';
+import { isAbsolute, join } from 'path';
 import { window, workspace, Uri, commands } from 'vscode';
 import { CommandPlugin } from './command';
+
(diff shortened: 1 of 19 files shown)


packages/engine/vscode/src/lib/filemanager.ts
```diff
@@ -1,49 +1,62 @@
...
+}
+
 export class FileManagerPlugin extends CommandPlugin implements MethodApi<IFileSystem> {
   constructor() {
     super(filSystemProfile);
   }
   /** Open the content of the file in the context (eg: Editor) */
   async open(path: string): Promise<void> {
-    const uri = Uri.parse(path);
+    const absPath = absolutePath(path);
```
yann300, line 21: you don't need to parse it anymore?

yann300: or is `Uri.file(absPath)` doing that?

yann300 (approved):

packages/engine/vscode/src/lib/filemanager.ts
GrandSchtroumpf: parse wasn't working as expected. Uri.file does (even for directories).

Issue #198, opened 2020-04-26
`activatePlugin` make sure it is activated before continuing


yann300:
```
await this.call('manager', 'activatePlugin', 'source-verification')
await this.call('activatePlugin', 'function1')
```
when the plugin `function1` is called, the plugin seems not to be fully activated, the call fails with `Error from IDE : No window attached to Iframe yet` .
In order to reproduce:
- tweak the `canActivate` of the IDE for returning always `true`
- create a javascript file in Remix IDE
```
(async () => {
    try {
        await remix.call('manager', 'activatePlugin', 'source-verification')
        const data = await remix.call('source-verification', 'fetch', '0x96d87AB604AEC7FB26C2E046CA5e6eBBB9D8b74D', 'ropsten')    
    } catch (e) {
        console.error(e.message)   
    }
})()
```

- run this script: `remix.exeCurrent()` from the remix terminal

GrandSchtroumpf:
The engine [wait for the plugin to be activated](https://github.com/ethereum/remix-plugin/blob/master/projects/engine/src/engine/engine.ts#L136) but when it comes to an external plugin (iframe, ...) it doesn't wait for the plugin client to be connected.
The main issue here is that there is no way for us to know if the plugin client is going to connect at all. For example maybe the plugin want the user to login before connecting to the IDE.
I **think** that it would goes inside the "timeout" after 10sec, but I still need to try it out.

I'll check how we can solve that.

PR #245, opened 2020-09-21
await for handshake

GrandSchtroumpf:
Make sure the plugin is ready to be called after activation.

### DIFF
--- a/packages/engine/core/src/lib/abstract.ts
+++ b/packages/engine/core/src/lib/abstract.ts
@@ -56,10 +56,10 @@ export class Plugin<T extends Api = any, App extends ApiMap = any> implements Pl
     this.profile.methods = methods
   }
 
-  activate() {
+  activate(): any | Promise<any> {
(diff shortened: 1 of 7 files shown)


packages/engine/core/src/lib/abstract.ts
```diff
@@ -56,10 +56,10 @@ export class Plugin<T extends Api = any, App extends ApiMap = any> implements Pl
     this.profile.methods = methods
   }
 
-  activate() {
+  activate(): any | Promise<any> {
     if (this.onActivation) this.onActivation()
```
yann300, line 60: shouldn't this `return` something?

GrandSchtroumpf: no it's a hook. It's not suppose to block the process.

yann300:
linked to https://github.com/ethereum/remix-plugin/issues/198

yann300 (approved):

PR #263, opened 2020-11-02
WIP: Remix Theia Plugin - Engine 

flantony:
This is intended as a basis for discussions regarding the implementation of the remix - Theia integration.


packages/engine/theia/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@remixproject:engine-theia",
```
GrandSchtroumpf, line 2: rename `@remixproject/engine-theia` with a "/" instead of ":"

packages/engine/theia/package.json
```diff
@@ -0,0 +1,4 @@
...
+  "version": "0.0.1"
```
GrandSchtroumpf, line 3: Use version "0.3.0" to keep the libs coherent

tslint.json
```diff
@@ -0,0 +1,65 @@
+{
```
GrandSchtroumpf, line 1: Use eslint for the linter, not tslint.
Best would be to remove the lib and rerun the schematic with eslint.
You can run `nx generate @nrwl/workspace:remove --projectName=engine-theia` to remove all files changes

flantony:
not relevant anymore

PR #268, opened 2020-11-09
change throw to console.warn

GrandSchtroumpf:


packages/plugin/iframe/src/lib/connector.ts
```diff
@@ -13,6 +13,9 @@ import {
 
 import { listenOnThemeChanged } from './theme'
 
+function warn() {
```
yann300, line 16: is that needed?

GrandSchtroumpf: ah sorry, I forgot to remove it

packages/plugin/iframe/src/lib/connector.ts
```diff
@@ -32,11 +35,11 @@ export class IframeConnector implements ClientConnector {
   /** Get messae from the engine */
   on(cb: (message: Partial<Message>) => void) {
     window.addEventListener('message', async (event: MessageEvent) => {
-      if (!event.source) throw new Error('No source')
+      if (!event.source) return console.warn('No source in message', event)
+      if (!event.data) return console.warn('No data provided in message', event)
```
yann300, line 39: At this point we should probably also filter messages that are supposedly not for us.
e.g which doesn't have an `action` property (and also other required properties)

GrandSchtroumpf: Alright

yann300 (approved):

PR #276, opened 2020-11-11
Theia extension example 

flantony:
Skeleton Theia Extension that fits into monorepo structure


flantony:
this is will be moved to separate repository

PR #280, opened 2020-11-23
add permission documentation

GrandSchtroumpf:


packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
 emptyPlugin.call('console', 'print', 'My message')
-```
\ No newline at end of file
+```
+
+--------------
+
+## Permission
+The Engine comes with a permission system to protect the user from hostile plugins.
+There are two level:
```
ryestew, line 152: There are two levels:

packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
+- **Global**: at the `PluginManager` level.
+- **Local**: at the `Plugin` level.
+
+### Global Permission
+Communication between plugins goes through the `PluginManager`'s permission system : 
+
+```typescript
+canActivatePlugin(from: Profile, to: Profile): Promise<boolean>
+```
+Used when a plugin attempts to activate another one. By default when a plugin "A" calls plugin "B", if "B" is not deactivated, "A" will attempt to active it before perform the call.
```
ryestew, line 162: Used when a plugin attempts to activate another one.  By default when plugin "A" calls plugin "B", if "B" is not deactivated, "A" will attempt to active it before performing the call.

packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
+```typescript
+canCall(from: Profile, to: Profile, method: string, message: string): Promise<boolean>
+```
+Used by a plugin to protect a method (see Local Permission below).
+
+**Tip**: Each method returns a `Promise`. It's good practice to ask the user's permission through a GUI.
+
+
+### Local Permission
+A plugin can protect some critical API by asking user permission:
```
ryestew, line 178: A plugin can protect some critical API by asking for user's permission:

packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
+
+```typescript
+askUserPermission(method: string, message: string): Promise<boolean>
+```
+This method will call the `canCall` method from the `PluginManager` under the hood with the right params.
+
+In this example a FileSystem plugin protect the `write` method :
```
ryestew, line 185: In this example, a FileSystem plugin protects the `write` method :

packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
+    if (!canCall) {
+      throw new Error('You do not have the permission to call method "canCall" from "fs"')
+    }
+  }
+}
+```
+
+### ⚠️ Warning ⚠️
+The permission system heavily relies on a queue of calls managed by the `Engine` and the property `currentRequest`.
+If you're calling a method from the plugin directly (without using the `Engine`) it will bypass the permission system. In this case you should not trust the result of `currentRequest`.
```
ryestew, line 201: If you're calling a method from the plugin directly (without using the `Engine`) it will bypass the permission system. In this case you should not trust the results of `currentRequest`.

( I just added changed result => results)

packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
+
+Example : 
+```typescript
+const fs = new FileSystemPlugin()
+const manager = new PluginManager()
+...
+fs.call('manager', 'activatePlugin', 'terminal')  // At this point `currentRequest` on manager is "fs"
```
ryestew, line 208: // At this point `currentRequest` in manager is "fs"

packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
+manager.deactivatePlugin('editor')  // This will fail
+```
+
+In the code above :
+1. call to "activatePlugin" enter the queue of the manager.
```
ryestew, line 213: 1. call to "activatePlugin" to enter the queue of the manager.

packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
+2. manager's `currentRequest` is "fs".
+3. manager calls its own `deactivatePlugin` method. 
+4. **as the call doesn't use the Engine, it doesn't enter in the queue**: `currentRequest` is still "fs".
```
ryestew, line 216: 4. **as the call doesn't use the Engine, it doesn't enter in the queue**: so `currentRequest` is still "fs".

packages/engine/core/README.md
ryestew: In this case you should not trust the results of currentRequest.  - I'm not sure this is clear / strong enough - here's an edit:
In this case,  the results of currentRequest may NOT be correct.

packages/engine/core/README.md
```diff
@@ -143,4 +143,75 @@ manager.activatePlugin(['empty', 'console'])
...
+5. `deactivatePlugin` checks the `currentRequest`, and it thinks that `fs` is trying to deactivate "terminal" and will fails.
```
ryestew, line 217: 5. `deactivatePlugin` checks the `currentRequest`. So now currentRequest incorrectly thinks that `fs` is trying to deactivate "terminal" and will not allow it.

ryestew: I am making this comment  based on my understanding in the call this morning.

ryestew: If I am understanding this correctly - maybe we need a title warning about this - like: "When currentRequest is Mistaken" or something like that.

ryestew (changes requested):

ryestew (approved):

PR #282, opened 2020-12-01
create Theia command plugin

flantony:


package.json
```diff
@@ -50,6 +50,7 @@
     "@types/vscode": "^1.47.0",
     "@typescript-eslint/eslint-plugin": "4.3.0",
     "@typescript-eslint/parser": "4.3.0",
+    "@theia/plugin": "latest",
```
GrandSchtroumpf, line 53: Please specify a version as you might not support the service forever, and external developers should know which version is supported at a given time.

GrandSchtroumpf (dismissed):

PR #294, opened 2020-12-14
Support global acquireTheiaApi function to webview connect 

bschwert:
Support global acquireTheiaApi function to connect plugin webview connector with engine

### DIFF
--- a/packages/plugin/webview/src/lib/connector.ts
+++ b/packages/plugin/webview/src/lib/connector.ts
@@ -11,6 +11,9 @@ import {
 } from '@remixproject/plugin'
 import { RemixApi, Theme } from '@remixproject/plugin-api';
 
+declare global {
+  function acquireTheiaApi(): any;
+}
 
 /** Transform camelCase (JS) text into kebab-case (CSS) */
 function toKebabCase(text: string) {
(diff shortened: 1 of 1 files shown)


packages/plugin/webview/src/lib/connector.ts
```diff
@@ -24,11 +27,16 @@ export class WebviewConnector implements ClientConnector {
   source: { postMessage: (message: any, origin?: string) => void }
   origin: string
   isVscode: boolean
+  isTheia: boolean
```
GrandSchtroumpf, line 30: Is there a reason to manage differently isVscode & isTheia in the `send` and `on` methods ?
If not could we only keep `isVscode` ?

packages/plugin/webview/src/lib/connector.ts
```diff
@@ -24,11 +27,16 @@ export class WebviewConnector implements ClientConnector {
...
 
   constructor(private options: PluginOptions<any>) {
     this.isVscode = ('acquireVsCodeApi' in window)
+    // TODO acquireTheiaApi is also available as acquireVsCodeApi in Theia. Can we check for global acquireVsCodeApi in both cases?
+    this.isTheia = acquireTheiaApi != undefined
```
GrandSchtroumpf, line 35: Could you write it like that : 
```typescript
// @todo(#295) check if we can merge this statement in `this.isVscode = acquireVsCodeApi !== undefined`
this.isVscode = ('acquireVsCodeApi' in window) || acquireTheiaApi !== undefined
if ('acquireVsCodeApi' in window) {
  this.source = window['acquireVsCodeApi']()
} else if (acquireTheiaApi !== undefined) {
  this.source = acquireTheiaApi()
} else {
  this.source = window.parent
}
```
Except if you need a specific behavior for `send` and `on` than vscode.

GrandSchtroumpf (changes requested):

GrandSchtroumpf:
closed in favor of #296

PR #307, opened 2021-01-05
dispose decorations

0mkara:


0mkara:
- [ ] dispose all decorations when dismissed
- [ ] use vscode theme color references

PR #311, opened 2021-01-13
createclient should return new pluginClient if none is provided as args in JS

bunsenstraat:
This only happens when using JS, not in TS. There you can pass anything or nothing. But in case it's nothing it should also return the new client, once it's been created if nothing is passed right? otherwise, you pass nothing, you get nothing back, you should get a new client instance.

### DIFF
--- a/packages/plugin/webview/src/lib/connector.ts
+++ b/packages/plugin/webview/src/lib/connector.ts
@@ -91,14 +91,14 @@ export const createClient = <
   C extends PluginClient<P, App> = any
 >(client: C): C & PluginApi<App> => {
   const c = client as any || new PluginClient<P, App>()
-  const options = client.options
+  const options = c.options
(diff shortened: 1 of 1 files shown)


packages/plugin/webview/src/lib/connector.ts
```diff
@@ -91,14 +91,14 @@ export const createClient = <
   C extends PluginClient<P, App> = any
 >(client: C): C & PluginApi<App> => {
```
GrandSchtroumpf, line 92: If you initialize in the parameter, no need for "c" anymore.
```typescript
createClient(client: C = new PluginClient<P, App>()) {
  const options = client.options;
  ....
  return client;
}
```

GrandSchtroumpf (dismissed):

yann300 (approved):

PR #313, opened 2021-01-15
Plugin API: support terminal plugin interface

dev-ion:
add new terminal api to **plugin-api** ([https://github.com/dev-ion/remix-plugin/tree/master/packages/api](url)).

```
export interface ITerminal {
  events: {} & StatusEvents;
  methods: {
    logTx(tx: Transaction): void;
    logHtml(msg: string): void;
  };
}
```

스크린샷 2021-01-15 오후 5 50 24
< example use of Klaytn Network (not Ethereum, but forked from Ethereum) ><br><br>
They can be used in plugins with custom network. (ex. Quorum Plugin)


GrandSchtroumpf:
I would recommend not to do that and provide a generic `log(message)` message instead, with a message that could be parse by the IDE engine and display if it match a specific pattern.
For me this kind of API is an open door to more `logX()` depending on the need of each plugin developer

GrandSchtroumpf:
@dev-ion I'm closing for house keeping. Let me know if you think it should be reopen

PR #317, opened 2021-02-02
API to return relative path

yann300:


bunsenstraat (commented):
;; on line 29

bunsenstraat (approved):

packages/engine/vscode/src/lib/filemanager.ts
```diff
@@ -1,91 +1,100 @@
...
+    return workspace.fs.createDirectory(uri)
   }
   /** Get the list of files in the directory */
   async readdir(path: string): Promise<string[]> {
-    const absPath = absolutePath(path);
-    const uri = Uri.file(absPath);
-    return workspace.fs.readDirectory(uri).then(data => data.map(([path]) => path));
+    const absPath = absolutePath(path)
+    const uri = Uri.file(absPath)
+    return workspace.fs.readDirectory(uri).then(data => data.map(([path]) => relativePath(path)))
```
yann300, line 83: I was about to ask, is that fine to do this : `data => data.map(([path]) => relativePath(path))` 
@0mkara @bunsenstraat

0mkara: I haven't checked that particular condition. Can't say without testing. Will do a test and let you know here.

PR #319, opened 2021-02-03
Themecolorfix

bunsenstraat:


packages/engine/vscode/src/lib/editor.ts
```diff
@@ -46,9 +46,8 @@ export class EditorPlugin extends CommandPlugin implements MethodApi<IEditor> {
       const start: Position = new Position(position.start.line, position.start.column);
       const end: Position = new Position(position.end.line, position.end.column);
       const newDecoration = { range: new Range(start, end) };
-      const codeThemeColor: string = extractColor(themeColor);
       this.decoration = window.createTextEditorDecorationType({
-        backgroundColor: new ThemeColor(codeThemeColor),
+        backgroundColor: new ThemeColor('editor.wordHighlightStrongBackground'),
```
0mkara, line 50: `themeColor` comes from variables used in the plugin. Why do you want to use a static colour from decoration?

0mkara (changes requested):

packages/engine/vscode/src/lib/editor.ts
bunsenstraat: There's an outstanding issue to remove the highlight color from the plugin API. If we'd require it plugins devs would have to know the names of these colors or provide a hex. Which is cumbersome imo. We can't expect that dependency. Plugins should not have to be rewritten to work in vscode ideally.

yann300: I would be fine to deprecate it and use the theme color..
just to be sure: `editor.wordHighlightStrongBackground` this variable is declared and defined by vscode right?

bunsenstraat: yes it's a list they provide. If you use new ThemeColor that is what you provide as a parameter. https://code.visualstudio.com/api/references/theme-color#editor-colors

yann300: oh ok thanks,  then we should deprecate this parameter and remove it from the public API

yann300 (approved):

0mkara (approved):

PR #338, opened 2021-02-23
fix for opening files and getting current file

bunsenstraat:
the window.activeTextEditor is often undefined when interacting with a plugin through clicks for example, because it is no longer active. This means using that to open a file or getting the current file will not work.

The solution is to find all the visibleTextEditors that have a document in it with a scheme of 'file' and open the file in with the 
TextDocumentShowOptions vielColunm property set to the viewcolumn of either the active editor or the first editor found that has a file in it
https://vshaxe.github.io/vscode-extern/vscode/TextDocumentShowOptions.html

to get the current file, the only option is to just to assume the first editor found with a file in it will be current, although you can have multiple columns open with files, when you lose focus there is no way of knowing which one is current.
it is important to note that if an editor has an unsaved empty untitled file in it, it is not viewed as editor with a file opened.

there is big caveat: if the plugin has focus in the same column as the files, for example when it is a mainpanel plugin, it will never find either activeeditor or texteditors opened. so for getcurrentfile it will never find the file. for openfile it will just open in the same column.

### DIFF
--- a/packages/engine/vscode/src/util/editor.ts
+++ b/packages/engine/vscode/src/util/editor.ts
@@ -10,12 +10,12 @@ export function getOpenedTextEditor() {
 
 export function getTextEditorWithDocumentType(type: string) {
   const editors: TextEditor[] = window.visibleTextEditors;
-  const fileeditor: TextEditor = editors.find(
+  const fileEditor: TextEditor = editors.find(
     (editor) =>
       editor.document &&
       editor.document.uri &&
       editor.document.uri.scheme &&
       editor.document.uri.scheme == type
   );
-  return fileeditor;
+  return fileEditor;
 }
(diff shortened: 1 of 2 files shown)


packages/engine/vscode/src/util/editor.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (!window.activeTextEditor) {
+    return getTextEditorWithDocumentType('file');
+  } else {
+    return window.activeTextEditor;
+  }
+}
+
+export function getTextEditorWithDocumentType(type: string) {
+  const editors: TextEditor[] = window.visibleTextEditors;
+  const fileeditor: TextEditor = editors.find(
```
0mkara, line 13: Let's use camelcase `fileEditor`

0mkara (changes requested):

yann300 (approved):

PR #350, opened 2021-03-29
Improve remix api security

yann300:


packages/engine/core/src/lib/connector.ts
```diff
@@ -1,5 +1,7 @@
 import type { ExternalProfile, Profile, Message } from '@remixproject/plugin-utils'
 import { Plugin, PluginOptions } from './abstract'
+import * as crypto from 'crypto'
```
GrandSchtroumpf, line 3: Everything in the engine-core was designed to be platform agnostic (browser / node). By adding a dependency on cypto (which is node specific) you might break the engine for the web.

bunsenstraat:
Can you describe the problem you want to solve?

packages/plugin/core/src/lib/connector.ts
```diff
@@ -34,19 +35,31 @@ export function isPluginMessage(message: any): message is Message {
  */
 export function connectClient(connector: ClientConnector, client: PluginClient = new PluginClient()) {
   let isLoaded = false
+  let account
+  connector.on(async ({ action, key, name, payload, id, signature, verifier, requestInfo, error }) => {
+    if (!signature && isLoaded) {
+      throw new Error('signature should have been set')
+    }
+    if (signature) {
+      const sign = fromRpcSig(signature)
```
Aniket-Engg, line 44: Not sure if this is necessary in our case: https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/modules/_signature_.md#const-fromrpcsig

packages/plugin/core/src/lib/connector.ts
```diff
@@ -34,19 +35,31 @@ export function isPluginMessage(message: any): message is Message {
...
+      const address = ecrecover(toBuffer(verifier), sign.v, sign.r, sign.s)
```
Aniket-Engg, line 45: This returns public key and it is compared with address: https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/modules/_signature_.md#const-ecrecover

packages/engine/core/src/lib/connector.ts
```diff
@@ -20,8 +22,16 @@ export interface PluginConnectorOptions extends PluginOptions {
...
+function newAccount () {
+  let privateKey: Buffer
+  do {
+    privateKey = crypto.randomBytes(32)
+  } while (!isValidPrivate(privateKey))
+  return privateKey
+}
 
 export abstract class PluginConnector extends Plugin {
+  protected account: any
```
Aniket-Engg, line 34: Can we give it a specific type instead of `any`?

packages/engine/core/src/lib/connector.ts
Aniket-Engg: what about using https://github.com/ethereumjs/ethereumjs-wallet/blob/master/docs/classes/wallet.md ?

PR #357, opened 2021-04-19
update webview css 

bunsenstraat:
trying to get the css to load in most conditions:
host       plugin.    load
https.     https.     https
http.       https.     https
https      http.       https
http        http.       http
localhost http.     http
localhost https.   http

now only an error will be thrown by https plugins on localhost, because they will try load https, fail, then load http
on our sites there will be no more errors

### DIFF
--- a/packages/plugin/webview/src/lib/connector.ts
+++ b/packages/plugin/webview/src/lib/connector.ts
@@ -10,6 +10,7 @@ import {
   isPluginMessage
 } from '@remixproject/plugin'
 import { IRemixApi, Theme } from '@remixproject/plugin-api';
+import axios from 'axios'
 
 
 /** Transform camelCase (JS) text into kebab-case (CSS) */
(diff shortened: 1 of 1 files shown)


packages/plugin/webview/src/lib/connector.ts
```diff
@@ -167,10 +168,33 @@ async function listenOnThemeChanged(client: PluginClient) {
...
+    }).catch(() => {
+        if(backupUrl) getLink().setAttribute('href', backupUrl);
+    });
+  }
+
   // If there is a url in the theme, use it
   const setLink = (theme: Theme) => {
     if (theme.url) {
-      getLink().setAttribute('href', theme.url.replace(/^http:/,"").replace(/^https:/,""))
+      const url = theme.url.replace(/^http:/, "protocol:").replace(/^https:/, "protocol:");
```
yann300, line 183: maybe that doesn't really apply in your case, but why not always load using https?

bunsenstraat: 1. because of localhost development. the https css is not there. It used to be that the theme url was always loaded, so it was not a problem but then running plugins on the web with a http host always failed. 2. and if the host is on http and plugin too we won't be sure the https css is there too.

yann300: I m not sure whether this can help the PR but we could also use the public http link `https://remix.ethereum.org/assets/css/...`
your solution is ok though too

bunsenstraat: yes that would help but from the perspective of an agnostic web view, the engine can be anything, for example vscode. the code from francois was really neutral, give me url, I will add it. you could argue that is the best solution and the host should take care of the urls being all https. ... I don't think my solution is very elegant.

yann300 (approved):

PR #360, opened 2021-04-27
Feature: Vscode extension API connector

0mkara:
# Objective
This PR adds the native plugin profile of `vscodeExtAPI`. This plugin is developed to expose vscode extension API to remix plugins.


0mkara:
Follow #361

PR #372, opened 2021-05-21
Workspaceapi

bunsenstraat:


packages/api/src/lib/file-system/file-panel/api.ts
```diff
@@ -9,7 +9,7 @@ export interface IFilePanel {
         customAction: (cmd: customAction) => void
     } & StatusEvents
     methods: {
-        getCurrentWorkspace(): { name: string, isLocalhost: boolean }
+        getCurrentWorkspace(): { name: string, isLocalhost: boolean, workSpacesPath: string }
```
yann300, line 12: could you explain why you need this params?

bunsenstraat: https://github.com/ethereum/remix-project/pull/990/files#diff-06867df237e883f50474f3dec9c1cd760265d4e8f1d20376f3807e6683c6582dR45

bunsenstraat: I know I shouldn't be public but what else ...

yann300: ah ok I see, in that case do you think we could return the absolute path of the workspace. like:
`{ name: string, isLocalhost: boolean, absolutePath: string }` .
Would this still work with dGit?

yann300 (approved):

PR #383, opened 2021-09-05
cancel method &  new queue

bunsenstraat:
This will allow

- canceling either all calls by a plugin to another plugin or a specific method called by a plugin to another plugin
```
// cancel all slowcalls in the queue from this client to someplugin
client.cancel('someplugin', 'slowcall')
// cancel all calls from this client  to someplugin
client.cancel('someplugin')
```
- controlling individual timeouts like this:
Normally all calls are queued in a plugin, this cannot be changed **. But using this method you can have big plugin timeout times but now cancel a call if it takes too long like in this example. The cancel method will reject the promise enabling displaying UI to retry or whatever...
```
// this slow call will be canceled after 1 second, even if the timeout of someplugin is 10secs.
try {
    setTimeout(() => {
        client.cancel('someplugin', 'slowcall')
    }, 1000)
    await client.call('someplugin', 'slowcall', data);
} catch (e) {
    console.log(e)
}
```

This will come in useful when you have a plugin that calls another plugin with a long timeout setting but you want to restrict certain calls to be shorter. For example in dGit react components try to connect to API's that are sometimes down while the others are up. It needs a long timeout to do IPFS cloning calls but shorter calls on for example checking if an IPFS node is up or not. 

This gets really important in situations where the Engine can not be restarted, for example in VSCode, and a queue might get saturated with long timeout calls effectively blocking everything until it gets restarted or everything times out. This also allows a plugin to clear any outstanding calls it has running from a previous setting although I think deactivation should actually clear the queue from its calls.

Of course all of this doesn't mean the actual call in the plugin that is being called is canceled too, for example a long API call will still go till the end in the plugin that is called, but the queue won't wait for it as it is canceled and rejected. This was always the case, also with timeouts, the originals calls would still go on.

** Looking at the core, each plugin has 1 queue, this is necessary because it has a property called currentrequest. this is used a lot so it needs to remain. But by controlling individual timeouts as with the example above blocking calls can be avoided. having more queues is not necessary this way. 
https://github.com/ethereum/remix-plugin/blob/57088bc6cf99bd08f1ec0e824c2b3c7560e995c5/packages/engine/core/src/lib/abstract.ts#L35

TODO:

- [ ] - when a plugin deactivates I think the queue should be cleared because it might be blocked by calls. I have seen this happen many time. The IDE you can reload but for example VSCode engine just gets stuck with long timeout calls. this could be a plugin option, because the dev always knows best. 
- [ ] - check support for service calls like plugin.servicecall ...
- [ ] - some more typescripty correctness
- [x] - unit tests to simulate canceling calls


packages/engine/core/src/lib/abstract.ts
```diff
@@ -77,48 +77,39 @@ export class Plugin<T extends Api = any, App extends ApiMap = any> implements Pl
     return this[method](...args)
   }
 
+  protected setCurrentRequest(request: PluginRequest) {
+    this.currentRequest = request
+  }
+
+  protected letContinue() {
+    delete this.currentRequest
+    this.queue = this.queue.filter((value) => {
```
yann300, line 86: old requests aren't removed from the queue?

packages/engine/core/src/lib/abstract.ts
```diff
@@ -77,48 +77,39 @@ export class Plugin<T extends Api = any, App extends ApiMap = any> implements Pl
...
-        }
-        clearTimeout(ref)
-        letcontinue()
-      })
-      // If there is only one request waiting, call it
-      if (this.requestQueue.length === 1) {
-        this.requestQueue[0]()
-      }
-    })
+      const queue = new PluginQueue(resolve, reject, request, method, this.options, args)
```
yann300, line 98: maybe `PluginQueueItem` is preferable here (first I thought that this object contains the list of pending requests)

packages/engine/core/src/lib/abstract.ts
```diff
@@ -230,6 +221,14 @@ export class Plugin<T extends Api = any, App extends ApiMap = any> implements Pl
     throw new Error(`Cannot use method "call" from plugin "${this.name}". It is not registered in the engine yet.`)
   }
 
+  /** Call a method of another plugin */
```
yann300, line 224: `cancel a request` ?

packages/engine/core/src/lib/engine.ts
```diff
@@ -136,6 +136,50 @@ export class Engine {
...
+  /**
+   * Cancels calls from a plugin to another
+   * @param caller The name of the plugin that calls the method
+   * @param path The path of the plugin that manages the method
+   * @param method The name of the method to be cancelled, if is empty cancels all calls from plugin
+   */
+  private async cancelMethod(caller: string, path: string, method: string) {
+    const target = path.split('.').shift()
+    if (!this.plugins[target]) {
+      throw new Error(`Cannot call ${target} from ${caller}, because ${target} is not registered`)
```
yann300, line 148: `Cannot cancel ${target} from ${caller},`

packages/engine/core/src/lib/engine.ts
```diff
@@ -136,6 +136,50 @@ export class Engine {
...
+    // Get latest version of the profiles
+    const [to, from] = await Promise.all([
+      this.manager.getProfile(target),
+      this.manager.getProfile(caller),
+    ])
+
+    // Check if plugin FROM can activate plugin TO
+    const isActive = await this.manager.isActive(target)
+    
+    if (!isActive) {
```
yann300, line 160: if `target` is not active you probably don't need to do anything..
at least `await this.manager.toggleActive(target)` doesn't seem needed

packages/engine/core/src/lib/abstract.ts
```diff
@@ -77,48 +77,39 @@ export class Plugin<T extends Api = any, App extends ApiMap = any> implements Pl
...
+      return value.canceled === false && value.timedout === false && value.finished === false
+    })
+    if (next) next.run()
+  }
+
   /** Add a request to the list of current requests */
   protected addRequest(request: PluginRequest, method: Profile<T>['methods'][number], args: any[]) {
     return new Promise((resolve, reject) => {
-      // Add a new request to the queue
-      this.requestQueue.push(async () => {
```
yann300, line 84: this can be removed:
`protected currentRequest: PluginRequest`

packages/engine/core/src/lib/abstract.ts
bunsenstraat: they are

packages/engine/core/src/lib/abstract.ts
bunsenstraat: not true, it is set
```
  protected setCurrentRequest(request: PluginRequest) {
    this.currentRequest = request
  }
```

PR #413, opened 2022-07-11
bump version

bunsenstraat:


Aniket-Engg (commented):
I think bumped version should be `0.3.9`

PR #433, opened 2023-02-13
Mvsrcfiles

bunsenstraat:
src/ was not included in output causing the package.json of each package to have incorrect references to index.js and d.ts because it would have lib/ but not src/lib so index.js would not be in src/ . This would cause errors in node when running for example remixd on a higher node version and will also cause problems when moving to a higher node version on the IDE as it won't compile because it can't find the index based on the package.json. the references get added by tsc btw, by default it's src unless you force it by setting the rootdir.


yann300:
I think you need to remove `package-lock.json`

bunsenstraat:
> I think you need to remove `package-lock.json`
it updates the lock to version 2 of npm. it was outdated.

yann300 (approved):

